PR #1, opened 2022-11-25
Merge changes done in napi-cs repo

vmoroz:
This PR merges the changes that I previously implemented in the [napi-cs dev branch](https://github.com/vmoroz/napi-cs/tree/dev).
Key changes:
- New object model. There is only `JSValue`. There are no other types yet. I am not sure if we need them. JavaScript has a single value type.
- Updated code generator to use the `JSModuleBuilder` and `JSClassBuilder`. The rest of the code gen is almost the same. In future we can add support for different methods signatures and TypeScript `.d.ts` generation.


Example/Example.cs
```diff
@@ -18,30 +18,35 @@ public class Example
...
-			Console.WriteLine("Example.Value.get()");
-			return NodeApi.String.From(this.value);
-		}
-		set
-		{
-			this.value = value.As<String>();
-			Console.WriteLine($"Example.Value.set({this.value})");
-		}
-	}
+  public void Hello(JSCallbackArgs args)
```
jasongin: I intentionally did not want the user code to have to deal with the callback args as a single object. The source generator should take care of marshalling between that JS args object and the C# method args.

So far the source generator that I wrote only goes as far as converting to `Value[]`, but my idea was to eventually support marshalling to C# methods with ordinary argument lists and argument types (with some limitations).

It's OK to merge this change for now, if there's a plan to fix it later.

Example/Example.csproj
```diff
@@ -5,6 +5,8 @@
 		<AssemblyName>example</AssemblyName>
 		<PublishAot>true</PublishAot>
 		<NativeLib>Shared</NativeLib>
+		<Nullable>enable</Nullable>
+		<AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
jasongin: Is unsafe allowed just for use of `stackalloc`? Ideally the user project shouldn't need to allow unsafe, even if the library does. If some generated code is unsafe, that should be moved into the library.

NodeApi/JSValue.cs
```diff
@@ -0,0 +1,94 @@
...
+  {
+    if (_scope.IsInvalid)
+    {
+      throw new InvalidOperationException("The value handle is invalid because its scope is closed");
+    }
+    return _handle;
+  }
+
+  public static JSValue Undefined => JSNativeApi.GetUndefined();
+  public static JSValue Null => JSNativeApi.GetUndefined();
```
jasongin: `GetNull()`

.editorconfig
```diff
@@ -1,10 +1,14 @@
...
 
 [*]
-indent_style = tab
-indent_size = 3
-tab_width = 3
-tab_size = 3
+indent_size = 2
+tab_width = 2
+tab_size = 2
+indent_style = space
```
jasongin, line 7: I personally prefer tabs, but I recognize I'm probably in the minority among C#/JS coders. So I will not fight it. :)

NodeApi.Module.temp.cs
```diff
@@ -0,0 +1,61 @@
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using static NodeApi.JSNative.Interop;
+using NodeApi.Examples;
+
+namespace NodeApi.Generated;
+
+public static class Module
```
jasongin: This is a temporary generated file that should not be checked in.

NodeApi/JSValueScope.cs
```diff
@@ -0,0 +1,49 @@
+using System;
+using static NodeApi.JSNativeApi.Interop;
+
+namespace NodeApi;
+
+public class JSValueScope : IDisposable
+{
+  private napi_env _env;
+  private bool _isDisposed = false;
+  [ThreadStatic] private static JSValueScope? t_current = null;
```
jasongin, line 10: I had been hoping this could be async-local instead of thread-static, so that it could be used across C# async calls. But I think you mentioned the  env handle is not valid from another thread? That's going to cause difficulty when a C# method called by JS does async work. We might need to set up a [synchronization context](https://learn.microsoft.com/en-us/archive/msdn-magazine/2011/february/msdn-magazine-parallel-computing-it-s-all-about-the-synchronizationcontext) so that C# code can switch to the correct thread when needed.

Anyway I think more investigation will be needed here to get to a good design. But this is not the highest priority, so it's fine the way it is for now.

NodeApi/JSNativeApi.Interop.cs
```diff
@@ -0,0 +1,635 @@
...
+        {
+          return libraryName == nameof(NodeApi) ? NativeLibrary.GetMainProgramHandle() : default;
+        });
+    }
+
+    //===========================================================================
+    // Specialized pointer types
+    //===========================================================================
+
+    public record struct napi_env(nint Handle);
```
jasongin, line 31: Nice, this is a great way to use `record` that I have not seen before.

NodeApi/JSNativeApi.Interop.cs
```diff
@@ -0,0 +1,635 @@
...
+    {
+      private readonly byte value;
+
+      public c_bool(bool value) => this.value = (byte)(value ? 1 : 0);
+
+      public static implicit operator c_bool(bool value) => new c_bool(value);
+      public static explicit operator bool(c_bool value) => value.value != 0;
+
+      public static c_bool True = new c_bool(true);
+      public static c_bool False = new c_bool(false);
```
jasongin, line 195: `readonly`

NodeApi/NodeApi.csproj
```diff
@@ -4,6 +4,8 @@
 		<RootNamespace>NodeApi</RootNamespace>
 		<AssemblyName>NodeApi</AssemblyName>
 		<IsPackable>true</IsPackable>
+		<Nullable>enable</Nullable>
```
jasongin: `Nullable` is already enabled for all projects via `Directory.Build.props`.

NodeApi/ModuleGenerator.cs
```diff
@@ -1,6 +1,9 @@
-﻿using System;
+using System;
+using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Text;
+using System.Xml.Linq;
```
jasongin, line 6: XML is not used.

NodeApi/JSPropertyDescriptorListOfT.cs
```diff
@@ -0,0 +1,188 @@
...
+    JSCallback callback,
+    JSPropertyAttributes attributes = JSPropertyAttributes.DefaultMethod)
+  {
+    Properties.Add(new JSPropertyDescriptor(name, callback, attributes));
+    return (TDerived)(object)this;
+  }
+
+  public TDerived AddMethod(
+    string name,
+    Func<TObject, Action> getMethod,
```
jasongin, line 83: I'm confused by the extra level of indirection for the method callbacks. Why wouldn't this parameter be just `Action<TObject, JSCallbackArgs>` or `Func<TObject, JSCallbackArgs, JSValue>` (or variations without args)? Is there some need for lazy initialization of the method callbacks?

NodeApi/JSPropertyDescriptorListOfT.cs
```diff
@@ -0,0 +1,188 @@
...
+    return (TDerived)(object)this;
+  }
+
+  public TDerived AddProperty(
+    string name,
+    JSCallback? getter,
+    JSCallback? setter,
+    JSPropertyAttributes attributes = JSPropertyAttributes.DefaultProperty)
+  {
+    Properties.Add(new JSPropertyDescriptor(name, getter, setter, attributes));
```
jasongin, line 31: For all callbacks from JS to C#, including property getters and setters and methods, I think there needs to be some code that catches any exception thrown by the C# callback and converts it to a JS error. (I had a TODO for that in the [`Object.InvokeCallback()`](https://github.com/jasongin/napi-dotnet/blob/main/NodeApi/Object.cs#L193) code deleted in this PR.)

jasongin (approved):
Looks good, I left a few comments but you don't have to resolve all the issues now since this is all work in progress.

I invited you as a collaborator on the repo so you should have permission to merge.

NodeApi.Module.temp.cs
vmoroz: Just FYI: the new code gen looks differently.

vmoroz: It seems that I copied it temporary and forgot to remove. Deleted.

.editorconfig
vmoroz: Ideally, we should follow some common conventions.
For example, https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md
Also, it would be nice to have automated code formatting,

Example/Example.cs
vmoroz: I agree with you that it should be ideal to have code gen that can be applied to an arbitrary set of arguments. We implemented similar approach for [ReactNative module methods](https://github.com/microsoft/react-native-windows/blob/main/vnext/Microsoft.ReactNative.Managed.IntegrationTests/TurboModuleTests.cs).
The `JSCallbackArgs` has a special role of being "untyped" fallback option that also can provide some advanced features such as `NewTarget`. In this context we can use a single `JSValue` argument for now. 
In future we should expand to a more complicated system and see if `JSCallbackArgs` can be removed.
E.g. in RNW we define custom serializers for custom user types. These are predefined [readers](https://github.com/microsoft/react-native-windows/blob/main/vnext/Microsoft.ReactNative.Managed/JSValueReader.cs) and [writers](https://github.com/microsoft/react-native-windows/blob/main/vnext/Microsoft.ReactNative.Managed/JSValueWriter.cs)
Our code gen is in this project: https://github.com/microsoft/react-native-windows/tree/main/vnext/Microsoft.ReactNative.Managed.CodeGen

Example/Example.csproj
vmoroz: Good point! Let me try to remove it.

NodeApi/JSPropertyDescriptorListOfT.cs
vmoroz: I am adding a TODO for now.

NodeApi/JSPropertyDescriptorListOfT.cs
vmoroz: It is just a sketch for how we can implement methods with custom signatures. 
We can use generic arguments `T1`, `T2`, etc. instead of `JSValue` in these signatures.
In the end they all call the `AddMethod` with `JSCallback` that accepts `JSCallbackArgs` and returns `JSValue`.

We can bypass the indirection, but then we must require the generated user code to use the unsafe API directly.

NodeApi/JSValue.cs
vmoroz: Fixed.

NodeApi/JSValueScope.cs
vmoroz: Let's do it in future iterations. I am not familiar with the async-local and not sure how the code can be used from different threads.

NodeApi/NodeApi.csproj
vmoroz: Good point! I did not notice it before and just copied what we had in the napi-cs. Fixed.

vmoroz:
> Looks good, I left a few comments but you don't have to resolve all the issues now since this is all work in progress.
> 
> I invited you as a collaborator on the repo so you should have permission to merge.

Thank you for the code review! I had addressed your feedback that required small changes. Let's address the feedback for bigger changes in separate PRs.

PR #2, opened 2022-11-30
Reorganize projects, create test runner project

jasongin:
- Split the main `NodeApi` project into `NodeApi.Generator` and `NodeApi.Runtime` projects.
   - I had to drop uses of `nameof`/`typeof` in the generator code that referenced runtime types, because the project reference caused the generator to silently fail. Unless there's a way to resolve that, we'll have to live with that limitation in the generator.
 - Add a `NodeApi.Test` project that dynamically generates XUnit test cases by scanning the `TestCases` directory and manages building the C# modules and running each JS test file. (The generated test cases appear in the VS test explorer.)
 - Move the existing `Example` project under `TestCases`. (It should be split into more distinct test cases later, but I have not modified it here.)

I decided to _build_ the test C# code as part of running the tests, _not_ as part of the normal solution build. The biggest reason for this is VS currently does not reload the source-generator into its build engine when the generator code changes, so iterating on the generator is confusing and annoying when you rebuild in VS and it doesn't use the new generator code, or even when you build from the command line but VS helpfully re-runs the stale generator.

I'm putting each test module in a separate directory, because each module has to be a separate C# project, and projects are simplest if they include all `.cs` files in the directory. The drawback of this approach is that building a test module is slow: it takes about 12s on my machine. So if we have a separate module for every test case, it's going to take a long time to run many tests. The test runner allows multiple `.js` files per C# module, where each `.js` file is a separate test case. So each C# test module gets built once but loaded separately for each test case.

JS test code can use the `assert` module to check for expected results. Any failed assertions or any output to `stderr` will cause the runner to consider the test failed. A log file from each test case (including stdout+stderr) is saved under `out/obj`.

Later I'd also like to integrate TypeScript into this test build process, so the C# generator can also generate TS type definitions for the C# APIs, and the test TS code can then reference those typedefs while getting compiled to JS when the test runs.


vmoroz (approved):
LGTM

PR #3, opened 2022-12-05
Implement code styling rules

vmoroz:
In this PR we are adopting the [coding style](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md) from the [dotnet/runtime](https://github.com/dotnet/runtime) project.
- The coding style rules are encoded in the `.editorconfig` file and can be checked by this command:
  - `dotnet format --severity info --verbosity detailed --verify-no-changes`
- Some issues can be fixed automatically by running:
  -  `dotnet format --severity info --verbosity detailed`
  
This PR has the following changes:
- `.editorconfig` file is copied from the `dotnet/runtime` project.
- Changes automatically created by the `dotnet format`.
- Changes to fix or ignore some automated code changes.
- Changes to fix `dotnet format` issues that cannot be done automatically.
- VS Code tasks and launch config created by running .Net tools.

Questions:
- Should we implement the pre-commit rule?
- Should we create a CI pipeline to check them?


.vscode/launch.json
```diff
@@ -0,0 +1,26 @@
...
+            // Use IntelliSense to find out which attributes exist for C# debugging
+            // Use hover for the description of the existing attributes
+            // For further information visit https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger-launchjson.md
+            "name": ".NET Core Launch (console)",
+            "type": "coreclr",
+            "request": "launch",
+            "preLaunchTask": "build",
+            // If you have changed target frameworks, make sure to update the program path.
+            "program": "${workspaceFolder}/out/bin/Debug/NodeApi.Test/net7.0/NodeApi.Test.dll",
+            "args": [],
```
jasongin: Launching the test DLL like this doesn't do anything, does it?

I don't think we have any need for a `launch.json` at this time. If the intention is to run tests, then the `test` task should be used. For debugging tests, use the debug feature of the ".NET Core Test Explorer" extension (recommended in `.vscode/extensions.json`), or VS Test Explorer.

.vscode/tasks.json
```diff
@@ -36,6 +29,30 @@
 				"reveal": "always"
 			},
 			"problemMatcher": "$msCompile"
+    },
+    {
+        "label": "publish",
```
jasongin: There is no need for a `publish` task here. The test modules are not published by this task because they aren't directly referenced by the project or solution. The test runner automatically builds and publishes the test modules before running them.

.vscode/tasks.json
```diff
@@ -2,24 +2,17 @@
...
-				"reveal": "silent"
-			},
-			"problemMatcher": "$msCompile"
-		},
+            "label": "build",
+            "command": "dotnet",
+            "type": "process",
+            "args": [
+                "build",
+                "${workspaceFolder}/Test/NodeApi.Test.csproj",
```
jasongin: I'd suggest building the `.sln` instead of a specific project. That way it will build the generator project (which is not referenced by the test project) and any other projects that are added to the solution.

Same for the `watch` task below.

.vscode/tasks.json
```diff
@@ -2,24 +2,17 @@
...
+                "/property:GenerateFullPaths=true",
+                "/consoleloggerparameters:NoSummary"
+            ],
+            "problemMatcher": "$msCompile"
```
jasongin, line 21: Keep the `group` and `presentation` sections deleted above. The `group` section sets this as the default build task so that it is run by the Ctrl+Shift+B shortcut. The `presentation` section prevents the build console window from being shown unless there are errors.

Test/TestCases/A/Another.cs
```diff
@@ -2,40 +2,42 @@
 
 namespace NodeApi.Examples;
 
+#pragma warning disable CA1822 // Mark members as static
+
```
jasongin, line 6: Should we suppress all code analysis for test modules?

Or we can suppress this specific warning during the build, by adding `NoWarn="CA1822"` property in `BuildTestModuleCSharp()`.

jasongin (commented):
> Should we implement the pre-commit rule?

Probably not necessary for now, if we assume developers are always using a tool (VS or VS Code) that enforces the rules in `.editorconfig`, and if there is a check in the PR build.

> Should we create a CI pipeline to check them?

Yes, I'm planning to set up PR and CI builds that will both run this check and unit tests.

Test/TestCases/A/Another.cs
vmoroz: `dotnet format` was actually changing instance methods in this class to static methods. It was causing the test to break.
Maybe in real code it should not be an issue because instance methods supposed to access the instance variables.
Let's start writing real tests before deciding to ignore this warning for all test modules.

.vscode/launch.json
vmoroz: Good point! I will remove it.

.vscode/tasks.json
vmoroz: The original tasks.json is restored.

.vscode/tasks.json
vmoroz: The original tasks.json is restored.

.vscode/tasks.json
vmoroz: The original tasks.json is restored.

vmoroz:
Replaced by PR #6 from this repo to pass CI.

PR #7, opened 2022-12-08
CLR hosting using C# AOT

jasongin:
- Code generation:
   - Generate two module-initialize methods. The one that supports unmanaged callers just forwards to the one that supports managed callers. This enables using the same generated initialization code in either AOT or hosted models.
 - Runtime:
   - Add `Hosting` subfolder/namespace to the `NodeApi` assembly. (I considered making it a separate project/assembly, and we still could, but for now it's simpler to keep it together.)
   - Add `HostFxr` class, which is a port of the minimum required CLR hosting APIs from `hostfxr.h` to C#.
   - Add `NativeHost` class, which (when AOT compiled) is responsible for hosting the CLR and initializing the `ManagedHost`.
   - Add `ManagedHost` class, which (when not AOT compiled) is responsible for loading/initializing additional .NET node modules.
 - Build:
   - Publish the `NodeApi` assembly as an AOT-compiled `.node` module that exports a native entrypoint for the `NativeHost`.
   - Generate and publish the `.runtimeconfig.json` file (that defines required CLR hosting parameters).
   - Also publish the _managed_ `NodeApi.dll` that contains the `ManagedHost` and managed `NodeApi` runtime for CLR-hosted node modules.
 - Test:
   - Enable running all the test cases in both AOT and hosted modes. The only difference is in how the module is imported.

This is working end-to-end now, though there are a couple important TODOs in the code:
  - Logic to locate hostfxr: It needs select the architecture matching the current process.
  - Loading the `ManagedHost` requires a full assembly name including version. This will break when we implement versioning and it's no longer `1.0.0.0`.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    // Don't show the "Required assets are missing..." dialog on startup. We'll manage tasks.json / launch.json manually.
+    "csharp.suppressBuildAssetsNotification": true
+}
--- a/Runtime/Hosting/NativeHost.cs
+++ b/Runtime/Hosting/NativeHost.cs
@@ -9,7 +9,7 @@ using static NodeApi.JSNativeApi.Interop;
 
 namespace NodeApi.Hosting;
 
-internal class NativeHost : IDisposable
+internal partial class NativeHost : IDisposable
 {
     private const string ManagedHostAssemblyName = nameof(NodeApi);
     private const string ManagedHostTypeName =
@@ -56,7 +56,7 @@ internal class NativeHost : IDisposable
         string runtimeConfigJsonPath = Path.Join(nodeApiHostDir, @"NodeApi.runtimeconfig.json");
         ////Console.WriteLine("CLR config: " + runtimeConfigJsonPath);
 
-        var managedHostAsssemblyPath = Path.Join(nodeApiHostDir, @"NodeApi.dll");
+        string managedHostAsssemblyPath = Path.Join(nodeApiHostDir, @"NodeApi.dll");
         ////Console.WriteLine("Managed host: " + managedHostAsssemblyPath);
 
         // Load the library that provides CLR hosting APIs.
@@ -139,7 +139,7 @@ internal class NativeHost : IDisposable
             &InitializeModule;
 
         const uint GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS = 0x00000004;
-        if (GetModuleHandleEx(
+        if (GetModuleHandleExW(
             GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS,
             (nint)functionInModule,
             out nint moduleHandle) == 0)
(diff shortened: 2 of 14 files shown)


Test/TestCases/A/hello.js
```diff
@@ -1,7 +1,9 @@
 const assert = require('assert');
 
-// Load the addon module.
-const test = require(process.env['TEST_NODE_API_MODULE_PATH']);
+// Load the addon module, using either hosted or native AOT mode.
+const dotnetModule = process.env['TEST_DOTNET_MODULE_PATH'];
+const dotnetHost = process.env['TEST_DOTNET_HOST_PATH'];
+const test = dotnetHost ? require(dotnetHost).require(dotnetModule) : require(dotnetModule);
```
vmoroz, line 6: >require(dotnetHost) [](http://example.com/codeflow?start=26&length=19)

I wonder if we can use similar technique to define multiple C# Node-API modules per DLL.
E.g.
   require('DLL path').load_module('module1')
then in another JS file:
   require('DLL path').load_module('module2')

Runtime/Hosting/NativeHost.cs
```diff
@@ -0,0 +1,172 @@
...
+        ////Console.WriteLine("Managed host: " + managedHostAsssemblyPath);
+
+        // Load the library that provides CLR hosting APIs.
+        HostFxr.Initialize();
+
+        // https://github.com/vmoroz/napi-cs/blob/dev/NodeApi.Sdk.CLR/Build/nativehost.cpp
+
+        // Initialize the CLR with configuration from runtimeconfig.json.
+        hostfxr_status status = hostfxr_initialize_for_runtime_config(
+            runtimeConfigJsonPath, initializeParameters: default, out _hostContextHandle);
```
vmoroz: >_hostContextHandle [](http://example.com/codeflow?start=70&length=18)

I wonder if wrapping it into a SafeHandle would make it safer/easier to use?

Runtime/Hosting/NativeHost.cs
```diff
@@ -0,0 +1,172 @@
...
+        CheckStatus(status, "Failed to inialize CLR host.");
+
+        try
+        {
+            // Get a CLR function that can load an assembly.
+            status = hostfxr_get_runtime_delegate(
+                _hostContextHandle,
+                hostfxr_delegate_type.load_assembly_and_get_function_pointer,
+                out load_assembly_and_get_function_pointer loadAssembly);
+            CheckStatus(status, "Failed to get CLR load-assembly function.");
```
vmoroz: >CheckStatus [](http://example.com/codeflow?start=12&length=11)

If it is an extension method, then we would not need to assign result to a variable.

Runtime/Hosting/NativeHost.cs
```diff
@@ -0,0 +1,172 @@
...
+        return filePathBuilder.ToString();
+    }
+
+    [LibraryImport("kernel32", SetLastError = true)]
+    private static partial uint GetModuleHandleExW(
+        uint flags,
+        nint moduleNameOrAddress,
+        out nint moduleHandle);
+
+    [DllImport("kernel32", SetLastError = true, CharSet = CharSet.Unicode)]
```
vmoroz: >DllImport [](http://example.com/codeflow?start=5&length=9)

I am just curious why LibraryImport did not work here.

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    // Don't show the "Required assets are missing..." dialog on startup. We'll manage tasks.json / launch.json manually.
```
vmoroz, line 2: >Required assets are missing... [](http://example.com/codeflow?start=23&length=30)

I think it was the dialog that generated a lot of changes to these files last time. :) Thanks for suppressing it!

vmoroz (approved):
:shipit:

Runtime/Hosting/NativeHost.cs
jasongin: `[LibraryImport]` doesn't support use of `StringBuilder` for getting string output.

In the latest revision I've changed this to `[LibraryImport]` using a stack-allocated `char*` instead.

Test/TestCases/A/hello.js
jasongin: We chatted about this. It's definitely possible, but would add complexity to module initialization and I'm not convinced it's needed. We can reconsider later.

PR #9, opened 2023-01-20
Add new tests and some code improvements.

vmoroz:
This PR has several changes:
- Allow unit tests to be in subfolders.
- Added new tests.
- Added `JSObject` and `JSArray` helper classses.
- Some code moved from `JSNativeApi` to `JSValue`, `JSObject`, and `JSArray`.


Runtime/JSObject.cs
```diff
@@ -0,0 +1,41 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace NodeApi;
+
+public struct JSObject : IEnumerable<(JSValue name, JSValue value)>, IEnumerable
```
jasongin: Should it be `IDictionary<JSValue, JSValue>`?

Runtime/JSArray.cs
```diff
@@ -0,0 +1,41 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace NodeApi;
+
+public struct JSArray : IEnumerable<JSValue>, IEnumerable
```
jasongin: Why not `IList<JSValue>` or `ICollection<JSValue>`?

jasongin: `IEnumerable<T>` extends `IEnumerable` so it's not necessary to declare the latter. Same for the other JS collections.

Runtime/JSArray.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    public JSValue this[int index]
+    {
+        get => _value.GetElement(index);
+        set => _value.SetElement(index, value);
+    }
+
+    public void Add(JSValue item) => _value["push"].Call(_value, item);
+
+    public JSArrayItemEnumerator GetEnumerator() => new(_value);
```
jasongin: I think `JSArrayItemEnumerator` can be internal, or even a private nested class.

jasongin: Also I'd suggest naming it `JSArrayEnumerator`.

Runtime/JSArrayBuffer.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    public JSArrayBuffer(object? external, ReadOnlyMemory<byte> data)
+        => _value = JSValue.CreateExternalArrayBuffer(external, data);
+
+    public JSArrayBuffer(ReadOnlyMemory<byte> data)
+        => _value = JSValue.CreateExternalArrayBuffer(null, data);
+
+    public Span<byte> Data => _value.GetArrayBufferInfo();
+
+    public int ByteLength => _value.GetArrayBufferInfo().Length;
```
jasongin: ArrayBuffers always use bytes as the element type, so I think just `Length` would be good.

I assume we'll add separate typed array structs later.

Runtime/JSException.cs
```diff
@@ -14,4 +18,43 @@ public unsafe JSException(string message)
     {
         ErrorInfo = new JSErrorInfo(message, JSStatus.GenericFailure);
     }
+
+    [DoesNotReturn]
+    public static void Fatal(string message, string location)
+        => napi_fatal_error(location, NAPI_AUTO_LENGTH, message, NAPI_AUTO_LENGTH);
```
jasongin: If two parameters are specified, then will the call be ambiguous due to the default values in the overload below?

Runtime/JSObject.cs
```diff
@@ -0,0 +1,41 @@
...
+        set => _value.SetProperty(name, value);
+    }
+
+    public void Add(JSValue name, JSValue value)
+        => _value.SetProperty(name, value);
+
+    public void Add((JSValue Key, JSValue Value) entry)
+        => _value.SetProperty(entry.Key, entry.Value);
+
+    public JSObjectPropertyEnumerator GetEnumerator() => new(_value);
```
jasongin: Similar to `JSArrayEnumerator`, I think `JSObjectEnumerator` can be internal, or private nested.

Test/TestCases/node-addon-api/object/delete_property.cs
```diff
@@ -23,7 +23,7 @@ private static JSValue DeletePropertyWithLatin1StyleString(JSCallbackArgs args)
     {
         JSValue obj = args[0];
         JSValue key = args[1];
-        return obj.DeleteProperty(JSNativeApi.CreateStringLatin1(key.GetValueStringLatin1()));
+        return obj.DeleteProperty(JSValue.CreateStringLatin1(key.GetValueStringLatin1()));
```
jasongin, line 26: Should there be a `JSString` type?

jasongin (approved):

Runtime/JSArray.cs
vmoroz: The main goal here was to mimic the [`List.Enumerator`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.enumerator?view=net-7.0) struct that allows enumerate `List` without extra memory allocations. I can make it a nexted class, but it must stay public to avoid memory allocations.

Runtime/JSObject.cs
vmoroz: Good point! I can add it in a follow up PR. Here I was focused only on `IEnumerable`. It would be still required for `IDictionary`. I just need to change the tuple to `KeyValuePair`.

Runtime/JSArray.cs
vmoroz: Good point! I can add it in a follow up PR.

Runtime/JSArrayBuffer.cs
vmoroz: Good point! I just started to work on this class and it is just a draft. I will change it. I must finish the `JSException` first.

Runtime/JSException.cs
vmoroz: Oops... It is still work in progress. I will see if I can remove all this "drafty"

Runtime/JSObject.cs
vmoroz: Please see my other comment: I need it to be public to avoid memory allocation as it is done in `List.Enumerator`.

Test/TestCases/node-addon-api/object/delete_property.cs
vmoroz: Good catch! I will change it.

Runtime/JSArray.cs
jasongin: Ahh that makes sense. I would prefer a nested `JSArray.Enumerator` class then, to follow the same pattern as `List` and avoid polluting the namespace.

Runtime/JSObject.cs
jasongin: I'd suggest nested `JSObject.Enumerator` then.

Test/TestCases/node-addon-api/object/delete_property.cs
jasongin: It doesn't need to be done in this PR.

vmoroz: We do not have `JSString` yet. I can change it later when we add it.

Runtime/JSException.cs
vmoroz: I removed this overload.

Runtime/JSArrayBuffer.cs
vmoroz: Changed to `Length`.

Runtime/JSArray.cs
vmoroz: It is removed.

vmoroz: I have added the `IList<JSValue>` - it is incomplete for now. We should see if we need all methods.

Runtime/JSObject.cs
vmoroz: I have added the `IDictionary<JSValue, JSValue>`. Some properties/methods are not implemented yet.

PR #10, opened 2023-01-22
Support multiple module instances with AssemblyLoadContext 

jasongin:
- Use `AssemblyLoadContext` in `ManagedHost` to load each module instance into a separate context.
   - I don't know of any equivalent solution for AOT modules. It might be just a limitation we have to accept/document.
 - Add test case to verify multiple instances and static data isolation
 - Add `global.json` to ensure use of correct .NET version during development.
 - Fix test builder selection of MSBuild version to match current runtime version.
 - Rename test subfolder `A` to `napi-dotnet`, to indicate test cases developed for this project that aren't ported from `node-addon-api`.


vmoroz (approved):
LGTM

PR #11, opened 2023-01-24
Add JSExport attribute

jasongin:
- Add `JSExportAttribute`.
 - Update source generation of module initialization code to support both `[JSModule]` and/or `[JSExport]` attributes.
    - See doc-comments on both of those attribute classes explaining their usage.
 - Add/update test cases for various export scenarios.


Runtime/JSExportAttribute.cs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /// <summary>
+    /// Exports an item as a property of the module exports object, with an explicit JavaScript
+    /// property name.
+    /// </summary>
+    /// <param name="name">Name of the item as exported to JavaScript.</param>
+    public JSExportAttribute(string name)
+    {
+        ArgumentException.ThrowIfNullOrEmpty(name, nameof(name));
```
vmoroz: >nameof(name) [](http://example.com/codeflow?start=51&length=12)

nit: it is not needed in C# 11. The name is deduced automatically from the first parameter.

Runtime/JSExportAttribute.cs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /// <summary>
+    /// Gets the name of item as exported to JavaScript, or null if the JavaScript name is
+    /// auto-generated by camel-casing the .NET name.
+    /// </summary>
+    /// <remarks>
+    /// Names must be unique among a module's exports. Duplicates will result in a build error.
+    /// <para/>
+    /// Use the name "default" to create a default export.
```
vmoroz, line 71: >Use the name "default" to create a default export. [](http://example.com/codeflow?start=8&length=50)

It seems non-obvious. What if someone wants to export a member with name "default"? It would be better to use a dedicated property or attribute.

vmoroz (approved):
:shipit:

Runtime/JSExportAttribute.cs
jasongin: > What if someone wants to export a member with name "default"?

That is exactly what a "default" export is: literally just an export that is named "default". Then JS/TS has some syntactic sugar for working with exports that match that naming convention.

I don't expect default exports to be widely used from C#, because typically there would be a JS wrapper for any published APIs anyway. I had initially created a separate `bool IsDefault` property on the attribute, but decided it was redundant and  unlikely to be used anyway.

Runtime/JSExportAttribute.cs
jasongin: Updated.

PR #12, opened 2023-01-28
Generate adapters for arbitrary parameters

jasongin:
This enables a cleaner more natural syntax for exporting methods, constructors, and properties that use only primitive types. See changes to the test C# code in which `JSCallbackArgs` and `JSValue` parameter and return types are replaced with C# types.

This is just a start, as it only handles basic conversions that can be done with a simple cast to/from `JSValue`. I'll extend it later to handle classes, structs, interfaces, delegates, etc.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,9 @@ jobs:
       if: ${{ always() }} # Run this step even when there are test failures
 
     - name: Check formatting
-      run: dotnet format --no-restore --severity info --verbosity detailed --verify-no-changes
+      run: |
+        dotnet tool install -g dotnet-format --version "7.*" --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet7/nuget/v3/index.json
+        dotnet format --no-restore --severity info --verbosity detailed --verify-no-changes
       if: ${{ always() }} # Run this step even when there are build failures
 
     # TODO: Publish packages
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,8 +72,7 @@ jobs:
       if: ${{ always() }} # Run this step even when there are test failures
 
     - name: Check formatting
-      run: dotnet tool install -g dotnet-format --version "7.*" --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet7/nuget/v3/index.json
-        dotnet format --no-restore --severity info --verbosity detailed --verify-no-changes
+      run: dotnet format --no-restore --severity info --verbosity detailed --verify-no-changes
       if: ${{ always() }} # Run this step even when there are build failures
 
     # TODO: Publish packages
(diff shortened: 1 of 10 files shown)


Generator/AdapterGenerator.cs
```diff
@@ -0,0 +1,298 @@
...
+using Microsoft.CodeAnalysis;
+
+namespace NodeApi.Generator;
+
+/// <summary>
+/// Generates C# <-> JavaScript adapter code for C# members exported to JavaScript.
+/// </summary>
+internal class AdapterGenerator : SourceGenerator
+{
+    private const string AdapterPrefix = "__";
```
vmoroz: >private const string AdapterPrefix = "__"; [](http://example.com/codeflow?start=4&length=42)

I am just curious why to prefix all types and variables with the double underscore? Can we just put them all in a special namespace?

vmoroz (approved):
:shipit:

.github/workflows/build.yml
```diff
@@ -75,6 +71,12 @@ jobs:
         reporter: dotnet-trx
       if: ${{ always() }} # Run this step even when there are test failures
 
+    - name: Check formatting
+      run: |
+        dotnet tool install -g dotnet-format --version "7.*" --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet7/nuget/v3/index.json
```
vmoroz: >dotnet-format [](http://example.com/codeflow?start=31&length=13)

How is it different from the built-in `dotnet format`?

jasongin: I was having trouble with the `dotnet format` tool: when it ran online it reported issues that were not reported when I ran locally. Furthermore some of the issues were just wrong: like it suggested making a method static when the method used an instance member. I thought installing the latest version might help. But it didn't make a difference, so I can remove this line.

Generator/AdapterGenerator.cs
jasongin: Types are not prefixed, only generated method names and variables in those methods.

Adapter method names include the full type name of the adapted method to avoid conflicts when multiple classes have the same method name. But they don't really need the extra `__` prefix since they are in a separately generated class.

I had considered generating the adapter methods in the same class as the target methods. That would require the class to be marked `partial`, similar to how `[LibraryImport]` generated methods work. Then the `__` or similar prefix or suffix would be needed. I didn't choose that approach because I didn't want to require modifying the original code.

The variables in generated code are prefixed with `__` to avoid any conflicts with the original method parameter names. in case the user's method has parameters named `args` or `result`. That prefix is a convention for generated code, and `[LibraryImport]` generated code does the same thing.

jasongin: Updated.

.github/workflows/build.yml
jasongin: Removed the dotnet-format install step.

PR #13, opened 2023-02-02
Generate TS type definitions for C# module exports

jasongin:
- Add `TypeDefinitionsGenerator` class
   - Generate type definitions for exported classes, properties, and methods.
   - Convert/reformat doc-comments for exported items.
 - Update some test JS code to reference the generated type definitions.

For now the TS generator only handles C# primitive types; anything else gets represented in TS as `any`. As I add support for exporting method parameters with other specific or complex types, I plan to implement the TS definitions at the same time.


vmoroz (approved):
LGTM

PR #14, opened 2023-02-06
Generate marshalling code for classes and structs

jasongin:
With these changes it is now possible to export methods with `class` and `struct` parameter types and return types, and have those values automatically marshalled from/to `JSValue`. Class instances are wrapped and marshalled by reference, as you'd expect.

C# `struct` types may now be tagged with `[JSExport]`, and struct instances are marshalled by value in attempt to preserve the C# value-type semantics. That means:
  - Structs do not use `napi_wrap`; when passing an instance across the C#-JS boundary the object properties are (shallowly) copied.
  - There is no link or lifetime relationship between a `struct` instance in C# and a corresponding object in JS.
  - When you construct an exported `struct` object in JS, a C# instance is _not_ immediately created.
  - When you construct an exported `struct` instance in C#, a JS instance is _not_ immediately created.
  - When you invoke an instance method on a `struct` object from JS, the JS object is copied into a newly-instantiated C# `struct` instance before the method is invoked.
      - This can be inefficient, so heavy use of `struct` methods from JS is not recommended. Instead, exported structs should generally be used as a convenient way to pass in/out a temporary bundle of related properties.

Overview of changes:
 - Add `JSStructBuilder<T>`, a parallel to `JSClassBuilder<T>` that supports the marshal-by-value semantics of structs.
 - Add a `JSContext` class that keeps track of constructor functions and class instance wrappers to facilitate the automatic marshalling.
 - Implement module lifetime semantics so that the module class and `JSContext` are disposed when the module is unloaded.
 - Update generated conversion code for parameters and return values to handle classes, structs, and nullables. 
 - Generate code to convert from/to `JSValue` to/from each exported `struct` type by copying/converting all properties.
 - Add direct JSValue conversion operators for nullable primitive types.
 - Update type definitions generator to include exported structs.
 - Add tests for above. (There's some test code for collections, but collections are not implemented yet in the generator.)

### DIFF
--- a/Runtime/JSNativeApi.cs
+++ b/Runtime/JSNativeApi.cs
@@ -459,12 +459,49 @@ public static partial class JSNativeApi
     public static JSValue Wrap<T>(T obj) where T : class
         => ObjectMap.GetOrCreateObjectWrapper(obj);
 
+    /// <summary>
+    /// Attaches an object to a JS wrapper.
+    /// </summary>
+    /// <param name="wrapper">The JS wrapper value, typically the 'this' argument to a class
+    /// constructor callback.</param>
+    /// <param name="value">The object to be wrapped.</param>
+    /// <returns>The JS wrapper.</returns>
     public static unsafe JSValue Wrap(this JSValue wrapper, object value)
     {
         GCHandle valueHandle = GCHandle.Alloc(value);
-        napi_wrap(Env, (napi_value)wrapper, (nint)valueHandle, new napi_finalize(&FinalizeGCHandle), nint.Zero, null).ThrowIfFailed();
+        napi_wrap(
+            Env,
+            (napi_value)wrapper,
+            (nint)valueHandle,
+            new napi_finalize(&FinalizeGCHandle),
+            nint.Zero,
+            null).ThrowIfFailed();
         return wrapper;
     }
+
+    /// <summary>
+    /// Attaches an object to a JS wrapper.
+    /// </summary>
+    /// <param name="wrapper">The JS wrapper value, typically the 'this' argument to a class
+    /// constructor callback.</param>
+    /// <param name="value">The object to be wrapped.</param>
+    /// <param name="wrapperWeakRef">Returns a weak reference to the JS wrapper.</param>
+    /// <returns>The JS wrapper.</returns>
+    public static unsafe JSValue Wrap(this JSValue wrapper, object value, out JSReference wrapperWeakRef)
+    {
+        GCHandle valueHandle = GCHandle.Alloc(value);
+        napi_ref weakRef;
+        napi_wrap(
+            Env,
+            (napi_value)wrapper,
+            (nint)valueHandle,
+            new napi_finalize(&FinalizeGCHandle),
+            nint.Zero,
+            &weakRef).ThrowIfFailed();
+        wrapperWeakRef = new JSReference(weakRef, isWeak: true);
+        return wrapper;
+    }
+
 
     public static object? Unwrap(this JSValue thisValue)
     {
--- a/Runtime/JSNativeApi.cs
+++ b/Runtime/JSNativeApi.cs
@@ -114,6 +114,13 @@ public static partial class JSNativeApi
     public static unsafe bool IsNull(this JSValue value)
         => value.TypeOf() == JSValueType.Null;
 
+    public static unsafe bool IsNullOrUndefined(this JSValue value) => value.TypeOf() switch
+    {
+        JSValueType.Null => true,
+        JSValueType.Undefined => true,
+        _ => false,
+    };
+
     public static unsafe bool IsBoolean(this JSValue value)
         => value.TypeOf() == JSValueType.Boolean;
 
@@ -382,7 +389,7 @@ public static partial class JSNativeApi
     }
 
     public static JSValue CallMethod(this JSValue thisValue, JSValue methodName)
-    => thisValue.GetProperty(methodName).Call(thisValue);
+        => thisValue.GetProperty(methodName).Call(thisValue);
 
     public static JSValue CallMethod(this JSValue thisValue, JSValue methodName, JSValue arg0)
         => thisValue.GetProperty(methodName).Call(thisValue, arg0);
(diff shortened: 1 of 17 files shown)


Runtime/JSValue.cs
```diff
@@ -249,6 +249,17 @@ public static unsafe JSValue CreateBigInt(int signBit, ReadOnlySpan<ulong> words
     public static implicit operator JSValue(ulong value) => CreateNumber(value);
     public static implicit operator JSValue(float value) => CreateNumber(value);
     public static implicit operator JSValue(double value) => CreateNumber(value);
+    public static implicit operator JSValue(bool? value) => value.HasValue ? GetBoolean(value.Value) : JSValue.Null;
```
vmoroz: >JSValue.Null [](http://example.com/codeflow?start=103&length=12)

Somewhere else in code we map C# null concept to JSValue.Undefined. 
I believe the JSValue.Undefined is a much better match than JSValue.Null. The JSValue.Undefined is the default value for JSValue.
When we return napi_value from a function the Node-API considers C++/C# null to be the same as napi_undefined.

vmoroz: >JSValue(bool? value) [](http://example.com/codeflow?start=36&length=20)

I wonder if we can use a generic instead.

Runtime/JSValue.cs
```diff
@@ -272,6 +288,22 @@ public static unsafe JSValue CreateBigInt(int signBit, ReadOnlySpan<ulong> words
     public static explicit operator string(JSValue value) => value.GetValueStringUtf16();
     public static explicit operator char[](JSValue value) => value.GetValueStringUtf16AsCharArray();
     public static explicit operator byte[](JSValue value) => value.GetValueStringUtf8();
+    public static explicit operator bool?(JSValue value) => value.IsNullOrUndefined() ? null : value.GetValueBool();
```
vmoroz: >bool? [](http://example.com/codeflow?start=36&length=5)

Can we use generic?

Runtime/JSNativeApi.cs
```diff
@@ -114,6 +114,13 @@ public static unsafe bool IsUndefined(this JSValue value)
     public static unsafe bool IsNull(this JSValue value)
         => value.TypeOf() == JSValueType.Null;
 
+    public static unsafe bool IsNullOrUndefined(this JSValue value) => value.TypeOf() switch
```
vmoroz, line 123: >IsNullOrUndefined [](http://example.com/codeflow?start=30&length=17)

In JavaScript it is usually called 'Nullish': [https://developer.mozilla.org/en-US/docs/Glossary/Nullish](https://developer.mozilla.org/en-US/docs/Glossary/Nullish)

Runtime/JSValue.cs
jasongin: I don't agree. If I call a C# method from JS, and the C# method returns `null`, I would expect to get a `null` value back in JS. Only `void` methods in C# (or property setters) should return `undefined`. Or maybe other situations where the value is truly uninitialized or unknown. I don't see anywhere else that's mapping C# `null` to JS `undefined`.

Runtime/JSNativeApi.cs
jasongin: Sometimes, but that term may be unfamiliar to many C# developers, or they may be confused about whether an empty string or other "falsy" values are also "nullish". So I prefer the more explicit name.

Runtime/ObjectMap.cs
```diff
@@ -0,0 +1,184 @@
...
+using static NodeApi.JSNativeApi.Interop;
+
+namespace NodeApi;
+
+/// <summary>
+/// Tracks JS constructors and instance JS wrappers for exported classes, enabling
+/// .NET objects to be automatically wrapped when returned to JS, and re-wrapped as needed
+/// if the (weakly-referenced) JS wrapper has been released.
+/// </summary>
+internal static class ObjectMap
```
jasongin: Instead of being `static`, this class needs to support multiple instances so that each instantiation of a module can track objects separately. That's necessary to resolve the `multi_instance` test failure.

Runtime/JSValue.cs
jasongin: Updated. It's a _little_ more concise.

Runtime/JSValue.cs
```diff
@@ -272,10 +283,27 @@ public static unsafe JSValue CreateBigInt(int signBit, ReadOnlySpan<ulong> words
...
+    public static explicit operator double?(JSValue value) => ValueOrNull(value, value => value.GetValueDouble());
 
     public static explicit operator napi_value(JSValue value) => value.GetCheckedHandle();
     public static implicit operator JSValue(napi_value handle) => new(handle);
 
     public static explicit operator napi_value(JSValue? value) => value?.Handle ?? new napi_value(nint.Zero);
     public static implicit operator JSValue?(napi_value handle) => handle.Handle != nint.Zero ? new JSValue?(new JSValue(handle)) : null;
+
+    private static JSValue ValueOrNull<T>(T? value, Func<T, JSValue> convert) where T : struct
+        => value.HasValue ? convert(value.Value) : JSValue.Null;
```
vmoroz, line 305: >convert [](http://example.com/codeflow?start=28&length=7)

Is it possible to use the existing conversions like this: `(T)value.Value`

jasongin: No, because the cast conversion is not defined for all possible `T`.

Runtime/ObjectMap.cs
```diff
@@ -0,0 +1,184 @@
...
+    /// </summary>
+    /// <param name="wrapper">JS object passed as the 'this' argument to the constructor callback
+    /// for <see cref="JSNativeApi.DefineClass"/>.</param>
+    /// <param name="obj">New or existing instance of the class to be wrapped.</param>
+    /// <returns>The JS wrapper.</returns>
+    internal static unsafe JSValue InitializeObjectWrapper<T>(JSValue wrapper, T obj) where T : class
+    {
+        GCHandle valueHandle = GCHandle.Alloc(obj);
+        napi_ref result;
+        napi_wrap(
```
vmoroz: >napi_wrap [](http://example.com/codeflow?start=8&length=9)

Can we use or add method from/to JSNativeApi.

jasongin: Yes. I'll add a `JSNativeApi.Wrap()` overload that returns the weak reference. The current one does not.

vmoroz (approved):
:shipit:

Runtime/JSValue.cs
jasongin: We'll postpone this for now; I opened https://github.com/users/jasongin/projects/1/views/1?pane=issue&itemId=19832782 to track it for later.

Runtime/ObjectMap.cs
jasongin: Updated.

PR #15, opened 2023-02-10
Generate adapter code for arrays and typed arrays

jasongin:
- Generate adapter code for arrays: C# `T[]` <-> JS `T[]` for any `T` that can be converted including primitives, structs, classes, etc.
   - Unfortunately array data must be copied because it's impossible to create a C# array over external memory.
 - Add `JSTypedArray<T>` value type.
 - Generate adapter code for typed arrays: C# `Memory<T>` <-> JS TypedArrays, where `T` is one of the valid typed-array element types, for example `Memory<int>` <-> `Int32Array`.
   - Typed arrays can be passed back and forth without copying the data, with some careful marshalling and lifetime management -- see `JSTypedArray.MemoryManager`.
 - Generate TS type definitions for arrays, typed arrays, and collections.

(Runtime adapter code is not yet generated for collections; I'm planning to work on that next. Collections should be adaptable without copying, by leveraging JS Proxy objects.)

### DIFF
--- a/Generator/AdapterGenerator.cs
+++ b/Generator/AdapterGenerator.cs
@@ -21,6 +21,7 @@ internal class AdapterGenerator : SourceGenerator
 
     private readonly Dictionary<string, ISymbol> _adaptedMembers = new();
     private readonly Dictionary<string, ITypeSymbol> _adaptedStructs = new();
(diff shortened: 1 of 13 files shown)


vmoroz (approved):
:shipit:

PR #16, opened 2023-02-10
Enable default for JSValue struct.

vmoroz:
Add default `JSValue` constructor that uses the default initialization of `JSValue` to zeroes.
The `default` `JSValue` is logically the same as `JSValue.Undefined`.

### DIFF
--- a/Runtime/JSValue.cs
+++ b/Runtime/JSValue.cs
@@ -9,30 +9,28 @@ namespace NodeApi;
 public readonly struct JSValue
 {
     private readonly napi_value _handle;
+    private readonly JSValueScope? _scope;
 
-    public JSValueScope Scope { get; }
+    public readonly JSValueScope Scope =>
+        _scope ?? JSValueScope.Current ?? throw new InvalidOperationException("No current scope");
 
-    public JSValue(JSValueScope scope, napi_value handle)
+    public JSValue() { }
+
+    public JSValue(napi_value handle) : this(JSValueScope.Current, handle)
     {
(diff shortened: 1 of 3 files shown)


Runtime/JSValue.cs
```diff
@@ -9,30 +9,28 @@ namespace NodeApi;
...
     }
 
-    public JSValue(napi_value handle)
+    public JSValue(JSValueScope? scope, napi_value handle)
     {
-        if (handle.Handle == nint.Zero)
+        if (handle.Handle != nint.Zero)
         {
-            throw new ArgumentException($"{nameof(handle)} must not be null.");
+            ArgumentNullException.ThrowIfNull(nameof(scope));
```
jasongin: `nameof(scope)` will never be null - delete `nameof()`.

jasongin: Is it valid to have a 0 handle and non-null scope?

jasongin:
Should we have similar default semantics for the other value types like `default(JSObject)` and `default(JSArray)`?

vmoroz:
> Should we have similar default semantics for the other value types like `default(JSObject)` and `default(JSArray)`?

We expect `JSObject` and `JSArray` always match their types, but there are no good defaults for these types.

jasongin (approved):

Runtime/JSValue.cs
vmoroz: Yes, it is valid - it will return Undefined in this case too.

PR #17, opened 2023-02-13
Generate adapter code for enumerables and lists

jasongin:
- Add `JSProxy` type for creating JS `Proxy` objects with C# handler callbacks.
 - Add `JSIterable` type for adapting JS iterable objects to/from `IEnumerable<T>`
 - Add `JSArray` support for adapting JS arrays to/from `IReadOnlyCollection<T>`, `ICollection<T>`, `IReadOnlyList<T>`, `IList<T>`
 - Use proxies to avoid copying collection items when passing from C# to JS.
 - Use `JSContext` to track strong references to proxy handlers and weak references to collection wrappers.
 - Add lots of comments to C# code generation.
 - Add finalizer to `JSReference` to release the JS reference when the C# object is GC'd.
    - When JS iterable or array is passed to C#, the marshaller avoids copying the data using a collection adapter. The adapter implements the C# collection interface by making calls to the JS value. It would be very unnatural to require C# code to explicitly dispose the collection adapter, especially since it is only referenced via an interface like `IList<T>`. I normally avoid relying on finalizers, but this seems like an appropriate use case.

The following collection interfaces are not yet implemented, though I plan to support them later using the same approach:
  - `IReadOnlySet<T>`, `ISet<T>` (project as JS `Set`)
  - `IReadOnlyDictionary<TKey, TValue>`, `IDictionary<TKey, TValue>` (project as JS `Map`)

Non-generic collections (`IEnumerable`, `ICollection`, `IList`, `IDictionary`) will require a different approach, using reflection, because the element conversions cannot be generated at compile time when the type is unknown. But I have no immediate plans to work on those because they aren't used as often anymore.


Runtime/JSReference.cs
```diff
@@ -64,11 +89,21 @@ protected virtual void Dispose(bool disposing)
     {
         if (!IsDisposed)
         {
+            // The reference handle should be deleted regardless of the `disposing` argument.
+            napi_status status = napi_delete_reference(_env, _handle);
+
+            // Ignore errors when finalizing.
             if (disposing)
             {
-                napi_delete_reference(_env, _handle).ThrowIfFailed();
```
vmoroz, line 69: >napi_delete_reference [](http://example.com/codeflow?start=16&length=21)

We are only allowed to call this function from the JS thread. Calling it from the finalizer thread will cause a crash because this API is not thread-safe.

Runtime/JSProxy.cs
```diff
@@ -0,0 +1,192 @@
...
+/// <summary>
+/// Enables creation of JS Proxy objects with C# handler callbacks.
+/// </summary>
+public readonly partial struct JSProxy
+{
+    private readonly JSValue _value;
+    private readonly JSValue _revoke;
+
+    public static explicit operator JSProxy(JSValue value) => new(value);
+    public static implicit operator JSValue(JSProxy arr) => arr._value;
```
vmoroz: >arr [](http://example.com/codeflow?start=52&length=3)

It seems to be a wrong parameter name.

Runtime/JSReference.cs
jasongin: Oh, well that's going to be a challenge. I assume the same is true for `napi_get_reference_value` and other things. I guess we'll need to set up a way to dispatch some calls to the JS thread. And then some C# test code that uses marshalled collections from another thread, to detect this kind of issue.

Runtime/JSNativeApi.cs
```diff
@@ -502,17 +502,38 @@ public static unsafe JSValue Wrap(this JSValue wrapper, object value, out JSRefe
         return wrapper;
     }
 
+    public static bool TryUnwrap(this JSValue thisValue, out object? value)
+    {
+        napi_status status = napi_unwrap(Env, (napi_value)thisValue, out nint result);
+        if (status == napi_status.napi_invalid_arg)
```
vmoroz, line 517: >if (status == napi_status.napi_invalid_arg) [](http://example.com/codeflow?start=8&length=43)

What is special meaning for this error code in this context?

Runtime/JSNativeApi.cs
```diff
@@ -502,17 +502,38 @@ public static unsafe JSValue Wrap(this JSValue wrapper, object value, out JSRefe
...
-    public static object? Unwrap(this JSValue thisValue)
+    public static object Unwrap(this JSValue thisValue)
     {
         napi_unwrap(Env, (napi_value)thisValue, out nint result).ThrowIfFailed();
-        return GCHandle.FromIntPtr(result).Target;
+        return GCHandle.FromIntPtr(result).Target!;
     }
 
-    public static object? RemoveWrap(this JSValue thisValue)
+    public static bool RemoveWrap(this JSValue thisValue, out object? value)
```
vmoroz, line 544: >RemoveWrap [](http://example.com/codeflow?start=23&length=10)

Should we name it TryRemoveWrap?

Runtime/JSReference.cs
jasongin: I see `napi_threadsafe_function` (added after I left the N-API project), that looks like it can be used for this kind of dispatch. I think I'll make a separate PR to set up that dispatching framework, then incorporate it here afterward.

vmoroz: Right. The expectation was that the Dispose method is always called from the JS thread. This why it was always called if disposing == true.

Runtime/JSNativeApi.cs
jasongin: That is the error code returned if `thisValue` does not have any wrapped external value. I'll add comments.

Runtime/JSNativeApi.cs
jasongin: Maybe, but I didn't because I expect most callers will ignore the return value because they don't care if the remove failed because it was never wrapped or already removed. It's similar to how `bool ICollection<T>.Remove(T item)` is not named `TryRemove`.

Runtime/JSReference.cs
vmoroz: Could you remove the call from the finalizer in this PR? The references are deleted on napi_env shutdown. There will be no memory leak. But with this new code there will be a crash.

jasongin: > The expectation was that the Dispose method is always called from the JS thread.

I don't think that's a reasonable assumption. It will be common for some other C# class to have a `JSReference` member, then when that class is disposed it would dispose the reference, possibly from another thread.

As you have said, when we do more async / threading testing we're likely to uncover lots of issues like this. But I _also_ think we will be able to put enough smarts in the NodeAPI library so that developers won't have to worry about it in most cases. I was sort of hoping we could wait on tackling that for just a little while longer though.

jasongin: > Could you remove the call from the finalizer in this PR? The references are deleted on napi_env shutdown.

Yes, I can remove it for now to avoid blocking the rest of this PR.

vmoroz: I agree that we may implement a better behavior in future. I am just asking to fix the issue identified in this PR.

Runtime/JSArray.Proxy.cs
```diff
@@ -0,0 +1,176 @@
+using System.Collections.Generic;
+
+namespace NodeApi;
+
+public partial struct JSArray
```
vmoroz, line 5: >JSArray [](http://example.com/codeflow?start=22&length=7)

I just wonder why these proxies are part of JSArray? I would expect them to be completely independent classes/structs. There seemed to be no relationship with JSArray.

Runtime/JSArray.Enumerator.cs
```diff
@@ -8,17 +8,17 @@ public partial struct JSArray
 {
     public struct Enumerator : IEnumerator<JSValue>, IEnumerator
```
vmoroz, line 9: >Enumerator [](http://example.com/codeflow?start=18&length=10)

JSArray already implements all these interfaces. Should we rather change the JSArray to be a generic class?

vmoroz (approved):
:shipit:

Runtime/JSArray.Enumerator.cs
jasongin: No. `JSArray` implements `IEnumerable<T>`, that returns an instance of this `Enumerator` struct. I didn't change this much, other than make the field name consistent with other internal `Enumerator` structs.

Runtime/JSArray.Proxy.cs
jasongin: The `Proxy()` methods that take an `IList<T>` or `IReadOnlyList<T>` actually do use a `JSArray` as the proxy target. The `Proxy()` methods that take an `ICollection<T>` or `IReadOnlyCollection<T>` use a plain object instead of an array. I didn't have a better place to put those methods because there is no JS "collection" class.

Note there is also a `Proxy()` method on `JSIterable`, and later I'll make similar ones for `JSSet` and `JSMap` that proxy the corresponding interfaces. It's only the "collection" proxies that don't have a good home. They could go in `JSIterable.Proxy.cs` instead, because they are iterables.

PR #18, opened 2023-02-16
Add async/await support and improved JSContext

vmoroz:
The main goal of this PR is to introduce support for C# async/await.
To implement it we are adding a new class `JSSynchronizationContext`. 
It can be created in the beginning of C# async function to be captured by `await` to continue execution on it after returning from a background thread. See the `async_method.js` and `AsyncMethod.cs` test as an example.

The `JSSynchronizationContext` is based on the new `JSThreadSafeFunction` wrapper for `napi_threadsafe_function`.
The `napi_threadsafe_function` has a ref count for threads that use it. It is controlled by the Acquire/Release calls.
If we have any thread that still uses the `napi_threadsafe_function`, then the environment cannot be shutdown.
Thus, I could not find a way to have a pseudo-global `JSSynchronizationContext` which is kept in `JSContext` for a lifetime of a module. So far, the only solution that works is to explicitly create the `JSSynchronizationContext` along with `JSThreadSafeFunction` when needed. I hope it should be possible to do in the generated code around async methods in future.

Since I was playing with the `JSContext` where I planned to store the `JSSynchronizationContext`, this PR also contains `JSContext` improvements:
- `JSContext` is not global. It is created per module.
- `JSContext` wraps up `napi_env`.
- `napi_env` encapsulates `JSContext` as its instance data. Thus, we can easily go between `JSContext` and `napi_env`.
- Since `JSContext` is storing itself as an instance data, it must be created after we establish `JSValueScope`. 
- `JSContext.Current` implementation is based on the `JSValueScope.Current`. We use it to get `napi_env` and then retrieve its instance data.
- `napi_env` in `JSReference` is replaced by `JSContext`. We can check `JSContext.IsDisposed` to see if we are allowed to do any `napi_ref` operations. In some sense the `JSContext` to `JSReference` as the `JSValueContext` to `JSValue`.
- `JSContext` holds the Module and responsible for its disposing.
- A number of other changes to code gen and hosting classes to enable this `JSContext` design improvements.

### DIFF
--- a/Runtime/JSContext.cs
+++ b/Runtime/JSContext.cs
@@ -2,7 +2,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using static NodeApi.JSNativeApi;
-using static NodeApi.JSNativeApi.Interop;
+using napi_env = NodeApi.JSNativeApi.Interop.napi_env;
 
 namespace NodeApi;
 
@@ -72,7 +72,7 @@ public sealed class JSContext : IDisposable
 
     public JSContext(napi_env env)
     {
-        Initialize();
+        Interop.Initialize();
         _env = env;
         SetInstanceData(this);
     }
@@ -331,14 +331,14 @@ public sealed class JSContext : IDisposable
         {
             IsDisposed = true;
 
-            DisposeReferences(_objectMap);
-            DisposeReferences(_classMap);
-            DisposeReferences(_structMap);
-
             if (Module is IDisposable module)
             {
                 module.Dispose();
             }
+
+            DisposeReferences(_objectMap);
+            DisposeReferences(_classMap);
+            DisposeReferences(_structMap);
         }
 
         GC.SuppressFinalize(this);
--- a/Runtime/JSContext.cs
+++ b/Runtime/JSContext.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using static NodeApi.JSNativeApi;
+using static NodeApi.JSNativeApi.Interop;
 
 namespace NodeApi;
 
@@ -15,10 +17,7 @@ namespace NodeApi;
 /// </remarks>
 public sealed class JSContext : IDisposable
 {
-    public JSContext()
-    {
-        JSNativeApi.Interop.Initialize();
-    }
+    private readonly napi_env _env;
 
     // Track JS constructors and instance JS wrappers for exported classes, enabling
     // .NET objects to be automatically wrapped when returned to JS, and re-wrapped as needed
@@ -59,6 +58,24 @@ public sealed class JSContext : IDisposable
     /// the JS object is not a wrapper, rather the properties are copied by the marshaller.
     /// </remarks>
     private readonly ConcurrentDictionary<Type, JSReference> _structMap = new();
+
+    public object? Module { get; set; }
+
+    public bool IsDisposed { get; private set; }
+
+    public static explicit operator napi_env(JSContext context) => context._env;
+
+    public static JSContext Current
+      => GetInstanceData() is JSContext context
+        ? context
+        : throw new InvalidCastException("Context is not found in napi_env instance data.");
+
+    public JSContext(napi_env env)
+    {
+        Initialize();
+        _env = env;
+        SetInstanceData(this);
+    }
 
     /// <summary>
     /// Maps from JS class names to (strong references to) JS constructors for classes imported
@@ -317,12 +334,15 @@ public sealed class JSContext : IDisposable
             DisposeReferences(_objectMap);
             DisposeReferences(_classMap);
             DisposeReferences(_structMap);
+
+            if (Module is IDisposable module)
+            {
+                module.Dispose();
+            }
         }
 
         GC.SuppressFinalize(this);
     }
-
-    public bool IsDisposed { get; private set; } = false;
 
     private static void DisposeReferences<TKey>(
         ConcurrentDictionary<TKey, JSReference> references) where TKey : notnull
--- a/Runtime/JSReference.cs
+++ b/Runtime/JSReference.cs
@@ -18,21 +18,19 @@ namespace NodeApi;
 /// </remarks>
 public class JSReference : IDisposable
 {
-    private readonly napi_env _env;
+    private readonly JSContext _context;
     private readonly napi_ref _handle;
 
     public bool IsWeak { get; private set; }
 
     public JSReference(JSValue value, bool isWeak = false)
+        : this(napi_create_reference((napi_env)JSValueScope.Current, (napi_value)value, isWeak ? 0u : 1u, out napi_ref handle).ThrowIfFailed(handle), isWeak)
     {
-        _env = (napi_env)value.Scope;
-        napi_create_reference(_env, (napi_value)value, isWeak ? 0u : 1u, out _handle).ThrowIfFailed();
-        IsWeak = isWeak;
     }
 
     public JSReference(napi_ref handle, bool isWeak = false)
     {
-        _env = (napi_env)JSSimpleValueScope.Current;
+        _context = JSContext.Current;
         _handle = handle;
         IsWeak = isWeak;
     }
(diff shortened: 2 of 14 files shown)


Runtime/JSContext.cs
```diff
@@ -15,10 +17,7 @@ namespace NodeApi;
 /// </remarks>
 public sealed class JSContext : IDisposable
 {
-    public JSContext()
-    {
-        JSNativeApi.Interop.Initialize();
-    }
+    private readonly napi_env _env;
```
jasongin, line 20: Since `JSContext` also wraps a `napi_env`, should it be a subclass of `JSValueScope`? And perhaps named something like `JSModuleScope`?

Runtime/JSContext.cs
```diff
@@ -60,6 +59,24 @@ public JSContext()
...
+    public static explicit operator napi_env(JSContext context) => context._env;
+
+    public static JSContext Current
+      => GetInstanceData() is JSContext context
+        ? context
+        : throw new InvalidCastException("Context is not found in napi_env instance data.");
+
+    public JSContext(napi_env env)
+    {
+        Initialize();
```
jasongin: It took me a little while to figure out that this is referring to `JSNativeApi.Interop.Initialize()`? How about making this more explicit, or changing `using static NodeApi.JSNativeApi.Interop` to `using napi_env = NodeApi.JSNativeApi.Interop.napi_env` ?

Runtime/JSContext.cs
```diff
@@ -317,13 +334,16 @@ public void Dispose()
             DisposeReferences(_objectMap);
             DisposeReferences(_classMap);
             DisposeReferences(_structMap);
+
+            if (Module is IDisposable module)
```
jasongin, line 334: Probably the module should be disposed before the references above, because it may depend on them.

Runtime/JSReference.cs
```diff
@@ -89,16 +87,11 @@ protected virtual void Dispose(bool disposing)
     {
         if (!IsDisposed)
         {
-            // The reference handle should be deleted regardless of the `disposing` argument.
-            napi_status status = napi_delete_reference(_env, _handle);
```
jasongin, line 93: The correct dispose pattern is to delete unmanaged resources (like this handle) regardless of the `disposing` value. That value is only false when invoked by a finalizer, and there is no finalizer yet, so it doesn't really make a difference. And when we do have a finalizer we'll have to fix it to switch to the right thread. But it was at least closer to being correct before.

Runtime/JSThreadSafeFunction.cs
```diff
@@ -0,0 +1,272 @@
...
+{
+    private napi_threadsafe_function _tsfn;
+
+    public static explicit operator napi_threadsafe_function(JSThreadSafeFunction function) => function._tsfn;
+    public static implicit operator JSThreadSafeFunction(napi_threadsafe_function tsfn) => new(tsfn);
+
+    private JSThreadSafeFunction(napi_threadsafe_function tsfn) => _tsfn = tsfn;
+
+    // This API may only be called from the main thread.
+    public static unsafe JSThreadSafeFunction New(int maxQueueSize,
```
jasongin: Is there a reason this is not a constructor?

jasongin:
> I could not find a way to have a pseudo-global `JSSynchronizationContext` which is kept in `JSContext` for a lifetime of a module.

I might have some ideas, but I'll wait to experiment after this is checked in.

> I hope it should be possible to do in the generated code around async methods in future.

Certainly that can be done. I was waiting on handling async methods in the code generator until we got this issue worked out. The simple thing to do will be to only set up a sync context for exported async methods. However technically it should be done for sync methods also, in case those sync methods somewhere call an `async void` method that needs to return to the JS thread after an `await`. However with the current design I think creating a sync context for every sync method would be too expensive. But we'll try to optimize that.

Runtime/JSSynchronizationContext.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    public override void Send(SendOrPostCallback callback, object? state)
+    {
+        if (this == Current)
+        {
+            callback(state);
+        }
+        else
+        {
+            Post(callback, state);
```
jasongin: I checked a few other implementations of `SynchronizationContext.Send()`, and they all synchronously wait for the callback execution to complete. But this looks like it just sends the message without waiting. One way to fix that could be to wrap the callback in a lambda that sets a `ManualResetEvent` after the inner callback returned, then wait on that callback in the outer method. Or maybe you have another idea.

Or if `Send()` is not used anywhere, maybe this doesn't matter.

Runtime/JSThreadSafeFunction.cs
```diff
@@ -0,0 +1,272 @@
...
+        {
+            FunctionContext = functionContext,
+            Finalize = finalize,
+            JSCaller = jsCaller
+        };
+        GCHandle functionDataHandle = GCHandle.Alloc(functionData);
+        napi_status status = napi_create_threadsafe_function(
+                                 (napi_env)JSValueScope.Current,
+                                 (napi_value)jsFunction,
+                                 (napi_value)(JSValue)(asyncResource is JSObject obj ? obj : new JSObject()),
```
jasongin: > `(asyncResource is JSObject obj ? obj : new JSObject())`

Can be shortened to `(asyncResource ?? new JSObject())`

jasongin: Actually the doc says it's optional, so can it default to `undefined` or a null handle?

jasongin (approved):

Runtime/JSContext.cs
vmoroz: We can consider doing it in future. My current concern about subclassing `JSContext` from `JSValueScope` is that it could be used by `JSValue` to govern its lifetime, but it will be not correct. `JSContext` can survive across multiple calls as long as `napi_env` is not finalized. While the `JSValueScope` always must have a lifetime bound to current call stack. 
The only common part is that they both keep `napi_env`. It seems that a better option would be to have `JSValueScope` keeping a pointer to `JSContext` instead of `napi_value`.
We should also consider renaming `JSContext` to `JSModuleContext`.
I can do these changes in a follow up PR.

vmoroz:
> I might have some ideas, but I'll wait to experiment after this is checked in.

I am curious to see how it can be done. I feel that the Acquire/Release pattern is a correct one, I just cannot find the right place to call the Release method when we have a TSFN owned by `JSContext`. TSFN keeps a strong reference to `napi_env` in terms of classical ref count Ref/Unref. TSFN destructor calls Unref. Thus, `napi_env` cannot be deleted until we have a single TSFN. If `JSContext` references TSFN, then we effectively have a reference count loop. We must find a way to break this loop.

My only hope is that we can subscribe to the process shutdown event to initiate breaking the ref count loop by calling TSFN Release. But to keep it working correctly, we still need to Acquire/Release for each async call. It is needed for BG not to crash when we call `Post`. Though, we may consider an alternative mechanism to protect the `Post` calls.

Runtime/JSContext.cs
vmoroz: Good point! I had the same experience. Fixed per your suggestion.

Runtime/JSContext.cs
vmoroz: Good point! Moved the module dispose up.

Runtime/JSReference.cs
vmoroz: I have removed the `disposing` value check. We must be careful with calling `napi_delete_reference`. Unlike other unmanaged resources our code does not own it 100%. We are only allowed to call it from the main JS thread and only before `napi_env` started finalization. It deletes all `napi_ref` as a part of the finalization process. Thus, we must always check the `JSContext.IsDisposed` before we call `napi_delete_reference`.

Runtime/JSThreadSafeFunction.cs
vmoroz: Changed to constructor. I started with multiple `New` overloads as in the `node-addon-api`, but lately found that if I change order of parameters and use default values, then I can use a single `New` method.

Runtime/JSSynchronizationContext.cs
vmoroz: Good point! Changed to have a true synchronous call. I often use this `InvokeElsePost` pattern and did it here because I thought docs say it too. Re-reading it shows that they were not saying it. It was just a wishful thinking on my side. :)

Runtime/JSThreadSafeFunction.cs
vmoroz: Checked the code and confirmed that if we pass `null`, then `napi_create_threadsafe_function` creates a new `Object` for us. No need to create it ourselves. I wish I understand what this object is needed for...

PR #19, opened 2023-02-16
Use JSSynchronizationContext per JSContext

vmoroz:
In PR we implement holding a single `JSSynchronizationContext` per `JSContext`.
Previously it was not clear how we can align the `JSSynchronizationContext` with the `JSContext` lifetime.
It seems that calling `Ref` and `Unref` for the `JSThreadSafeFunction` is the right API to notify Node-API if we use the thread safe function (TSFN) or not.

We create TSFN in the `JSSynchronizationContext` and then immediately `Unref` it.
The new `JSAsyncScope` is added to call `JSSynchronizationContext` `OpenAsyncScope` and `CloseAsyncScope`.
It can be explicitly added to C# async functions, or we can add it in our generated code for async functions later.

The `JSThreadSafeFunction` is changed to a class, and we added there a `_refCount` for correct `napi_ref_threadsafe_function` and `napi_unref_threadsafe_function` calls. These functions do not support nested calls and do not keep internal ref count. Thus, we must use our own ref count.

Added new `JSContext.RunInMainLoop` method that allows running `Action` callback in the main loop for the module.
This method can be called from any thread and schedules asynchronous running of the callback in the associated `JSSynchronizationContext`.
The additional `allowSyncRun` parameter can be set to `true` to allow synchronous execution in case if we are already in our `JSSynchronizationContext` thread.

The `JSValueScope` is refactored to represent five different types of scopes:
- Root scope - we create it at the beginning and it is responsible for `JSContext` creation.
- RootNoContext scope - we create it at the beginning and it does not create `JSContext`.
- Callback scope - we must use it at the beginning of each callback.
- Handle scope - it is to wrap `napi_handle_scope`.
- Escapable scope - it is to wrap `napi_escapable_handle_scope`.

Changes to `JSValueScope` allowed to delete `JSSimpleValueScope` and `JSEscapableValueScope` classes and files.


Runtime/JSAsyncScope.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using System.Threading;
+
+namespace NodeApi
+{
+    /// <summary
+    /// Keeps JSSynchronizationContext queue alive until the end of the 
+    public class JSAsyncScope : IDisposable
```
jasongin: Missing end of sentence and `</summary>`.

jasongin: Can `JSAsyncScope` be a struct, to avoid a heap allocation?

Runtime/JSAsyncScope.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using System.Threading;
+
+namespace NodeApi
```
jasongin: nit: Please use file-scoped namespaces.

Runtime/JSAsyncScope.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        public void Dispose()
+        {
+            // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+            Dispose(disposing: true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
```
jasongin: If this class is marked `sealed`, or if it's a struct, then this can be non-virtual (faster).

Runtime/JSAsyncScope.cs
jasongin: Or can it be combined somehow with `JSValueScope`, which is already created and disposed for every callback?

jasongin (approved):

Runtime/JSAsyncScope.cs
vmoroz: No, I do not see how we can combine them. They have different lifetime:
- `JSValueScope` is attached to the same call stack.
- `JSAsyncScope` is spans across multiple call stacks: we create it in one task, and then dispose it eventually in one of the continuation tasks. Its goal is to say to `JSSynchronizationContext` when we start using it and when we release it. 
If we would combine the two, then `JSValue` created in the initial call stack would not be able to understand that the underlying `napi_value` is already destroyed because we would fail to Dispose it at the right time.

vmoroz: More complete comments are written,

vmoroz: Make the JSAsyncScope a struct.

Runtime/JSAsyncScope.cs
vmoroz: Fixed.

Runtime/JSAsyncScope.cs
vmoroz: Changed to struct.

Runtime/JSValueScope.cs
```diff
@@ -1,50 +1,103 @@
 using System;
+using System.Threading;
 using static NodeApi.JSNativeApi.Interop;
 
 namespace NodeApi;
 
-public class JSValueScope : IDisposable
+public enum JSValueScopeType { Handle, Escapable, Callback, Root, RootNoContext, }
+
+public sealed class JSValueScope : IDisposable
```
jasongin, line 9: Can `JSValueScope` be a struct also (now that there is no longer a class hierarchy)? That would save a heap allocation on every callback.

jasongin (approved):

Runtime/JSValueScope.cs
vmoroz: I would like it to be. It cannot be a `struct` because we reference it from `JSValue`. I want to explore if we can make it `ref struct`, but I want to do it in a follow up PR. This one already is grown too much. :)

PR #20, opened 2023-02-17
Generate adapter code for sets and maps

jasongin:
- Add `JSSet` type with 2-way proxy/adapter code.
 - Add `JSMap` type with 2-way proxy/adapter code.
 - Add `JSSymbol` type. Cache a reference to the iterator symbol that is used by collection proxies.
 - Implement `IEquatable<JSValue>` in all JS value types.
 - Improve formatting of generated code so that it doesn't form super long lines.


vmoroz (approved):
:shipit:

PR #21, opened 2023-02-17
Export C# enums

jasongin:
- Generate module registration code to define C# enums as a JS objects with readonly properties.
 - Generate TS type definitions for enums.


vmoroz (approved):
LGTM

PR #22, opened 2023-02-17
Convert dates

jasongin:
Add `JSDate` type with conversions to/from `DateTime`.

### DIFF
--- a/Runtime/JSDate.cs
+++ b/Runtime/JSDate.cs
@@ -17,17 +17,17 @@ public readonly struct JSDate : IEquatable<JSValue>
 
     public JSDate()
     {
-        _value = JSValue.Global["Date"].CallAsConstructor();
+        _value = JSContext.Current.Import("Date").CallAsConstructor();
     }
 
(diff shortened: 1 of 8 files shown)


Runtime/JSDate.cs
```diff
@@ -0,0 +1,83 @@
...
+    public static implicit operator JSValue(JSDate date) => date._value;
+
+    private JSDate(JSValue value)
+    {
+        _value = value;
+    }
+
+    public JSDate()
+    {
+        _value = JSValue.Global["Date"].CallAsConstructor();
```
vmoroz: >JSValue.Global["Date"] [](http://example.com/codeflow?start=17&length=22)

Should we consider caching these constructors in JSContext?

vmoroz (approved):
:shipit:

Runtime/JSDate.cs
jasongin: Oh, I forgot I already built a way to do that with `JSContext.Current.Import('Date')`. Updated.

PR #24, opened 2023-02-17
JSPromise and async methods

jasongin:
- Add `JSPromise` JS value type.
 - Generate adapters for async methods.
 - Generate type definitions for async methods.
    - There is some TS support for `CancellationToken`s, but they are not implemented yet in C#.
 
 Constructors for `JSPromise` allow for JS style callback-based promises:
 ```C#
return new JSPromise(async (resolve) =>
{
    await Task.Delay(50);
    resolve(JSValue.True);
});
```

If an async resolve callback is provided, a `JSAsyncScope` is automatically used.

It should also be possible to convert between a `Task` and `JSPromise`, but I have not done that.

### CODE REVIEW

File: Runtime/JSPromise.cs [hunk 1]
```diff
@@ -0,0 +1,165 @@
...
+    /// <remarks>
+    /// Any (sync or async) exception thrown by the callback will be caught and used as a promise
+    /// rejection error.
+    /// </remarks>
+    public JSPromise(AsyncResolve callback)
+    {
+        _value = JSValue.CreatePromise(out Deferred deferred);
+        async void AsyncCallback()
+        {
+            using var __asyncScope = new JSAsyncScope();
```
vmoroz: >__asyncScope [](http://example.com/codeflow?start=22&length=12)

Can we avoid using double underscore for the local variable names?

jasongin: Oh, that was a copy-paste mistake here. The double-underscore should only be used in generated code.

vmoroz (approved):
:shipit:

vmoroz:
It looks awesome!

PR #25, opened 2023-02-18
Custom marshal attribute

jasongin:
This is a proposal for a custom marshalling attribute and interface. Custom marshalling is not actually implemented yet in the code-generator. The design is mostly modeled after [System.Runtime.InteropServices.MarshalAsAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshalasattribute).

### DIFF
--- a/Runtime/JSMarshalAsAttribute.cs
+++ b/Runtime/JSMarshalAsAttribute.cs
@@ -1,0 +1,37 @@
+
+using System;
+
+namespace NodeApi;
+
+/// <summary>
+/// Specifies how .NET types are marshalled to and from JavaScript values.
+/// </summary>
+[AttributeUsage(
+    AttributeTargets.Property |
+    AttributeTargets.Parameter |
+    AttributeTargets.ReturnValue)]
+public sealed class JSMarshalAsAttribute : Attribute
+{
+    /// <summary>
+    /// Specifies how a .NET property, parameter, or return value is marshalled to
+    /// and from a JavaScript value.
+    /// </summary>
+    /// <param name="marshal">One of the available marshaling options.</param>
+    public JSMarshalAsAttribute(JSMarshal marshal)
+    {
+        Value = marshal;
+    }
+
+    /// <summary>
+    /// Gets the marshaling behavior indicated by the attribute.
+    /// </summary>
+    public JSMarshal Value { get; }
+
+    /// <summary>
+    /// When <see cref="Value" /> is <see cref="JSMarshal.CustomMarshaler" />, gets the type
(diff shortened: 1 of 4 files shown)


Runtime/JSMarshalAsAttribute.cs
```diff
@@ -0,0 +1,37 @@
+
+using System;
+
+namespace NodeApi;
+
+/// <summary>
+/// Specifies how .NET types are marshalled to and from JavaScript values.
+/// </summary>
+[AttributeUsage(
```
vmoroz, line 9: >AttributeUsage [](http://example.com/codeflow?start=1&length=14)

Should we also add fields?

Runtime/JSMarshalAsAttribute.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    public JSMarshal Value { get; }
+
+    /// <summary>
+    /// When <see cref="Value" /> is <see cref="JSMarshal.CustomMarshaler" />, gets the type
+    /// that handles marshalling. The type must implement <see cref="IJSMarshaler{T}" />,
+    /// where the type argument matches the type being of the property, parameter, or
+    /// return value the <see cref="JSMarshalAsAttribute" /> is applied to.
+    /// </summary>
+    public Type? CustomMarshalType { get; init; }
```
vmoroz, line 36: >Type? [](http://example.com/codeflow?start=11&length=5)

I wonder if we can use the new generic attributes instead. In that case the restriction that the type must inherit from IJSMarshaler can be written as an explicit restriction checked by the C# compiler.

vmoroz (approved):
:shipit:

Runtime/JSMarshalAsAttribute.cs
jasongin: I thought about that, but we'd still need a non-generic `JSMarshalAsAttribute` to support uses that don't specify a custom marshaller type. So I didn't think it was worth creating another attribute.

Runtime/JSMarshalAsAttribute.cs
jasongin: Later, if and when we support marshalling fields.

Fields are not supported now, but there's no reason they can't be marshalled in almost the same way as properties, along with an optimization for `const` fields. I don't think it's a priority now though.

jasongin:
I'm abandoning this PR because it's not a top priority at the moment, and was mainly meant as a starting point for discussion. I do still plan to implement this idea eventually. Since opening this PR I've done a lot more work related to marshalling, so I think there can be some improvements to this design when I bring it back.

PR #26, opened 2023-02-19
Marshal interfaces

jasongin:
- Define JS classes that represent C# interfaces.
 - Generate C# proxy classes for exported interfaces.
 - Generate type definitions for exported interfaces.
 - Add `IJSValue` interface, which facilitates conversion between all the different types that represent or proxy a JS value, and enables the marshaller to get the original JS value behind a C# proxy.
 - Implement more methods on `JSPromise`. Async interface method implementation required `then()`, but I went ahead and implemented all the other standard promise APIs.
 - Add extension methods for converting between C# tasks and JS promises.

When a C# object is marshaled to JS as an interface type, the interface type is represented in JS as a _class_. An interface class cannot be constructed from JS; it is only constructed by the C# marshalling code. (Defining interface properties with a class prototype is more efficient than defining the interface properties on every marshalled object.)

When a JS object is marshaled to C# as an interface type, it uses a C# proxy class that implements the interface and forwards calls to the JS object. A proxy class is generated for each exported interface type.


Runtime/IJSValue.cs
```diff
@@ -0,0 +1,23 @@
...
+/// <remarks>
+/// This interface is implemented by:
+///   - Value types that directly represent a specific type of JS value such as
+///     <see cref="JSObject" />, <see cref="JSArray" />, etc.
+///   - Collection proxy classes returned by <see cref="JSIterable.AsEnumerable()" /> and similar
+///     methods, that forward all calls to a JS collection object.
+///   - Interface proxy classes auto-generated by <see cref="JSExportAttribute" />, that forward
+///     all calls to a JS object that implements the interface.
+/// </remarks>
+public interface IJSValue : IEquatable<JSValue>
```
vmoroz: >IJSValue [](http://example.com/codeflow?start=17&length=8)

It means that now we must box and unbox JSValues. It goes against us being efficient and performant. Could we use an alternative solution with static abstract methods instead that would help us avoid boxing?

vmoroz (approved):
:shipit:

Runtime/IJSValue.cs
jasongin: > It means that now we must box and unbox JSValues

No, not normally.

`IJSValue` is meant to be used when you have some _already boxed_ object and you need to find out whether it represents a JS value. It doesn't change any of the existing patterns that avoid boxing. Does that make sense?

An example of where it's used is when marshalling interfaces. An object that implements an interface might be an actual C# class, or it might be a generated proxy class to a JS object that implements the interface, so the marshaller uses `IJSValue` to detect the latter so it can return the JS object directly instead of making an unnecessary reverse proxy.

I originally considered naming it `IJSValueAccessor` or something like that, and then only the interface proxies and collection proxies would implement it. But I thought it could be more broadly useful if all the JS value types implement it also.

vmoroz: I guess it makes sense then. I am not sure if the `IJSValue` is the best name for an interface for such purpose.

In one of my prototypes, I implemented [`ITypedValue`](https://github.com/vmoroz/napi-dotnet/blob/dev/vmoroz/prototypes/typed_model/Runtime/TypedModel/TypedModel.cs) that used to get `JSValue`. In that code the interface and custom attributes were used to generate boilerplate code for the partial structs and use them uniformly. It was a static interface to avoid boxing/unboxing. We could use the name `IJSValue` instead of `ITypedValue` and do the same.

Another concern is that this interface looks so "basic" that other developers may be tempted to use it and think that it is a good practice. E.g. to write extensions methods. I would rather have them writing extensions based on the static interfaces instead of the boxed value. I wonder if the marshaller can be smart enough to look for structs and then see if it has a static interface implemented.

jasongin: A static interface won't work for the same purpose because the generic type argument is required while the precise type of the object isn't known.

I see your point about how the interface can be confusing. I can remove it and use a different approach in the marshaller that doesn't rely on a public interface.

jasongin: Removed. It wasn't really needed, the code is simpler now.

PR #27, opened 2023-02-20
Dynamic load & invoke with emitted marshaling code

jasongin:
This is an experiment / proof of concept for an alternative way of calling C# APIs from JS, that doesn't rely on source generation at the time the C# code is built. It uses reflection to scan the APIs but then _emits IL code_ for the JS marshaling layer. That means that while the initial load may be slower, subsequent API calls are just as fast as the source-generation approach.

So far it works only for the simplest case: invoking static methods with primitive parameter types. (See the `assembly_load.js` test case.) It does not yet support all the more complex marshaling scenarios handled by the source generator, but all that is possible.

I can imagine this approach being used in a few ways:
 1. You're writing some quick (non-type-checked) JavaScript and you need to invoke some .NET API. You can just load the assembly from a file path and start calling APIs in it:
```JavaScript
const dotnet = require('node-api-dotnet');
const assembly = dotnet.loadAssembly('path/to/ExampleAssembly.dll');
const ExampleClass = assembly['Namespace.ExampleClass'];
ExampleClass.ExampleMethod(...args);
```
 2. You're writing TypeScript (or JS with type-checking) and you want intellisense and type checking for .NET API calls. So you run the (future) Node API type-definitions generator tool on the .NET assembly, and that outputs a `.d.ts` file along with a small corresponding `.js` assembly loader that makes it possible to `import` the .NET library and use it naturally from TypeScript code:
```bash
$ node-api-ts-gen "path/to/ExampleAssembly.dll"
Generated ExampleAssembly.js
Generated ExampleAssembly.d.ts
```
```TypeScript
import { ExampleClass } from './ExampleAssembly';
ExampleClass.ExampleMethod(...args); // This is type-checked!
```
 3. Eventually, somebody might create an npm package feed (or scope in the global feed) that is full of pre-generated and packaged type definitions and loader scripts for popular .NET libraries:
```bash
npm install "@node-api-dotnet/ExampleAssembly"
```

I don't think dynamic emitting of IL marshaling code at runtime would completely replace the use of build-time source generation, since the latter is still more efficient for applications that are developing both C# and JS code together. Anyway it's much easier to implement marshaling logic in the source generator first, then look at the IL code compiled from that to inform development of the dynamic loader. (The **ILSpy** tool can show a nice view of decompiled C# with inline IL.) Also, source generation is the only approach compatible with AoT compilation, since neither reflection nor IL emitting is supported in that environment.

### DIFF
--- a/Runtime/Hosting/AssemblyExporter.cs
+++ b/Runtime/Hosting/AssemblyExporter.cs
@@ -1,0 +1,336 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Emit;
+
+namespace NodeApi.Hosting;
+
+using static ManagedHost;
+
+[RequiresUnreferencedCode("Managed host is not used in trimmed assembly.")]
+[RequiresDynamicCode("Managed host is not used in trimmed assembly.")]
+internal class AssemblyExporter
+{
+    private readonly JSReference _assemblyObject;
+    private readonly Dictionary<Type, JSReference> _typeObjects = new();
+    private readonly AssemblyBuilder _assemblyBuilder;
+    private readonly ModuleBuilder _moduleBuilder;
+
+    public AssemblyExporter(Assembly assembly)
+    {
+        Assembly = assembly;
+        _assemblyObject = new JSReference(ExportAssembly());
+        _assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(
+            new AssemblyName("JS." + assembly.FullName),
+            AssemblyBuilderAccess.Run);
+        _moduleBuilder = _assemblyBuilder.DefineDynamicModule("JS");
+    }
+
+    public Assembly Assembly { get; }
+
+    public JSValue AssemblyObject => _assemblyObject.GetValue()!.Value;
+
+    private JSValue ExportAssembly()
+    {
+        Trace($"> AssemblyExporter.ExportAssembly({Assembly.FullName})");
+
+        // Reflect over public types in the loaded assembly.
+        // Define a property for each type, with a getter that lazily defines the class.
+
+        List<JSPropertyDescriptor> assemblyProperties = new();
+        void AddAssemblyTypeProperty(Type type, Func<Type, JSValue> getter)
+        {
+            assemblyProperties.Add(JSPropertyDescriptor.Accessor(
+                type.FullName!,
+                (_) => getter(type),
+                setter: null,
+                JSPropertyAttributes.Enumerable));
+        }
+
+        foreach (var type in Assembly.GetTypes().Where((t) => t.IsPublic))
+        {
+            if (type.IsClass)
+            {
+                if (type.IsAbstract && type.IsSealed) // static class
+                {
+                    AddAssemblyTypeProperty(type, ExportStaticClass);
+                }
+                else
+                {
+                    AddAssemblyTypeProperty(type, ExportClass);
+                }
+            }
+            else if (type.IsValueType)
+            {
+                if (type.IsEnum)
+                {
+                    AddAssemblyTypeProperty(type, ExportEnum);
+                }
+                else
+                {
+                    AddAssemblyTypeProperty(type, ExportStruct);
+                }
+            }
+        }
+
+        JSObject assemblyObject = new JSObject();
+        assemblyObject.DefineProperties(assemblyProperties);
+
+        Trace($"< AssemblyExporter.ExportAssembly() => [{assemblyProperties.Count}]");
+        return assemblyObject;
+    }
+
+    private JSValue ExportStaticClass(Type classType)
+    {
+        Trace($"> AssemblyExporter.ExportStaticClass({classType.FullName})");
+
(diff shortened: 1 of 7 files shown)


jasongin:
I haven't decided yet how to handle .NET namespaces. A cleaner alternative to the above syntax would be to just strip namespaces:
```JavaScript
const assembly = dotnet.loadAssembly('path/to/ExampleAssembly.dll');
assembly.ExampleClass.ExampleMethod(...args);
```

It is possible, but unusual, for a .NET assembly to have multiple public types with the same name in different namespaces. So there would have to be some way for JS code to disambiguate if necessary.

Note TypeScript supports but [discourages use of namespaces](https://www.typescriptlang.org/docs/handbook/modules.html#do-not-use-namespaces-in-modules).

Runtime/JSObject.cs
```diff
@@ -31,6 +31,11 @@ public void DefineProperties(params JSPropertyDescriptor[] descriptors)
         _value.DefineProperties(descriptors);
     }
 
+    public void DefineProperties(IReadOnlyCollection<JSPropertyDescriptor> descriptors)
```
vmoroz, line 34: >IReadOnlyCollection [](http://example.com/codeflow?start=33&length=19)

I am surprised that we cannot use ReadOnlySpan. Sigh...

jasongin: Because `JSPropertyDescriptor` is a `class` it can't be a member of a `ref struct`. And it has managed objects like delegates that can't be contained by a `ref struct` either.

jasongin:
> I haven't decided yet how to handle .NET namespaces.

After writing more example code for the doc, and thinking some more, I'm planning on the following:
 - Non-generic types with names that are unambiguous within their namespace will be exported directly as properties on the assembly object. That enables simpler more idiomatic JS syntax for most types, like `const { TypeA, TypeB } = dotnet.loadAssembly(assemblyPath)`.
 - Generic types with unambiguous names will have their base name without any type parameters exported, but the returned value will be unusable until converted to a generic type somehow, something like `const ExampleCollectionOfString = assembly.ExampleCollection.makeGenericType(dotnet.String)`
 - Both generic and non-generic types can be always be referenced by full name, including namespace and type arguments if applicable, e.g. `assembly['Namespace.AmbiguousClass']` or `assembly['Namespace.ExampleCollection<System.String>']`. Probably that would support primitive type aliases like `string` also.

I plan to use a JS Proxy to defer scanning all the types in an assembly and defining properties for all of them, since most of the time the JS code will only select a few types by name.

Runtime/Hosting/AssemblyExporter.cs
```diff
@@ -0,0 +1,336 @@
...
+        // TODO
+    }
+
+    private void EmitStaticPropertySet(TypeBuilder typeBuilder, PropertyInfo property)
+    {
+        // TODO
+    }
+
+    private static void EmitLoadAndConvertArg(
+        ILGenerator il,
```
vmoroz: >ILGenerator [](http://example.com/codeflow?start=8&length=11)

I wonder why not use the [https://learn.microsoft.com/en-us/dotnet/api/system.linq.expressions?view=net-7.0](https://learn.microsoft.com/en-us/dotnet/api/system.linq.expressions?view=net-7.0) instead.
I found that generating code this way is much simpler and less error prone. I used to use it in RNW before we switched to compile time code generation (see in the end of file): [https://github.com/microsoft/react-native-windows/blob/main/vnext/Microsoft.ReactNative.Managed/JSValueWriterGenerator.cs](https://github.com/microsoft/react-native-windows/blob/main/vnext/Microsoft.ReactNative.Managed/JSValueWriterGenerator.cs)

Runtime/JSObject.cs
vmoroz: Sorry, it was me ranting about the .Net limitations, not the actual change. I just wish we can use Snap in situations like this.
I guess in our code we must convert the JSPropertyDescriptor to ref struct as you did with the JSCallbackArgs.

vmoroz (approved):
:shipit:

Runtime/Hosting/AssemblyExporter.cs
jasongin: I'll experiment with that approach.

jasongin:
> :shipit:

I wasn't planning to check this in as it is now, as it's very incomplete... though I guess it wouldn't hurt.

Runtime/JSObject.cs
jasongin: > I guess in our code we must convert the JSPropertyDescriptor to ref struct

I think that can be done by converting each descriptor to `napi_property_descriptor` immediately when constructed rather than as a batch later. It would require changing the whole ClassBuilder pattern also.

I'm not working on it now.

Runtime/Hosting/AssemblyExporter.cs
jasongin: Update: switched to lambda expressions.

It's roughly the same number of lines of code (perhaps more verbose), but I think it is safer compared to using the emit API where it was very easy for a bug to result in invalid IL.

jasongin:
Abandoning this PR. I'll open a new PR where this is all more fully implemented.

jasongin:
https://github.com/jasongin/napi-dotnet/pull/32

PR #28, opened 2023-02-21
Optimize JS callbacks

jasongin:
- Change `JSCallbackArgs` from a `class` with heap-allocated args array to a `ref struct` with stack-allocated args span.
 - Fix an issue uncovered by the perf test (due to new timing or memory use patterns I guess?) in which the JS GC could collect the static class object that caused the `JSPropertyDescriptor` GC handle to be finalized even though JS code could still call the method afterward.

With these changes, performance on the edgejs scenario is 10-15% faster.

### DIFF
--- a/Runtime/JSCallbackArgs.cs
+++ b/Runtime/JSCallbackArgs.cs
@@ -1,5 +1,7 @@
 using System;
 using System.Runtime.InteropServices;
+using static System.Formats.Asn1.AsnWriter;
+using static System.Runtime.InteropServices.JavaScript.JSType;
 using static NodeApi.JSNativeApi.Interop;
 
 namespace NodeApi;
@@ -10,11 +12,30 @@ public readonly ref struct JSCallbackArgs
     private readonly napi_value _thisArg;
     private readonly ReadOnlySpan<napi_value> _args;
 
-    internal JSCallbackArgs(
-        JSValueScope scope, napi_value thisArg, ReadOnlySpan<napi_value> args, object? data = null)
+    internal unsafe JSCallbackArgs(JSValueScope scope,
+                                   napi_env env,
+                                   napi_callback_info callbackInfo,
+                                   Span<napi_value> args,
+                                   object? data = null)                    
     {
+        nint dataPointer;
+        napi_value thisArgHandle;
+        if (args.Length == 0)
+        {
+            napi_get_cb_info(env, callbackInfo, null, null, &thisArgHandle, &dataPointer)
+                .ThrowIfFailed();
+        }
+        else
+        {
+            fixed (napi_value* argv = &args[0])
+            {
+                nuint argc = (nuint)args.Length;
+                napi_get_cb_info(env, callbackInfo, &argc, argv, &thisArgHandle, &dataPointer)
+                    .ThrowIfFailed();
+            }
+        }
         _scope = scope;
-        _thisArg = thisArg;
+        _thisArg = thisArgHandle;
         _args = args;
         Data = data;
     }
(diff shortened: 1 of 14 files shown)


Runtime/JSValue.cs
```diff
@@ -10,31 +10,29 @@ namespace NodeApi;
 public readonly struct JSValue : IEquatable<JSValue>
 {
     private readonly napi_value _handle;
-    private readonly JSValueScope? _scope;
-
-    public readonly JSValueScope Scope =>
-        _scope ?? JSValueScope.Current ?? throw new InvalidOperationException("No current scope");
+    private readonly napi_env _scope;
```
vmoroz: >private readonly napi_env _scope; [](http://example.com/codeflow?start=4&length=33)

We effectively trading performance for safety. While it may seem a good approach, it goes against our ability using this framework in server environment.

I do not like this change. After that any programming error will result in memory corruption which will be every expensive or impossible to find. I spent too many weeks catching memory corruptions.

If we want speed, then we must find an alternative, but do not trade safety of the code.

Runtime/JSCallbackArgs.cs
```diff
@@ -1,51 +1,57 @@
+using System;
 using System.Runtime.InteropServices;
 using static NodeApi.JSNativeApi.Interop;
 
 namespace NodeApi;
 
-public class JSCallbackArgs
+public readonly ref struct JSCallbackArgs
```
vmoroz, line 7: >readonly ref struct [](http://example.com/codeflow?start=7&length=19)

This is a good change! I can see how it makes the perf 15% faster because we effectively avoid extra memory allocation per each call.

Runtime/JSCallbackArgs.cs
```diff
@@ -1,51 +1,57 @@
...
 {
-    private readonly JSValue[] _args;
+    private readonly Span<JSValue> _args;
```
vmoroz: >Span [](http://example.com/codeflow?start=21&length=4)

Should it be ReadOnlySpan?

Runtime/JSCallbackArgs.cs
vmoroz, line 7: >ref struct  [](http://example.com/codeflow?start=16&length=11)

Can we use 'struct' instead of 'ref struct'?

vmoroz:
Please do not merge this PR until we can find a better solution. Improving the performance for each interop call is very important, but we must not regress on the code safety.

Runtime/JSCallbackArgs.cs
jasongin: It has to be `ref struct` so that it can have a `Span` member, to support stack-allocation of the args array.

Runtime/JSCallbackArgs.cs
jasongin: Yes.

vmoroz:
Please see the fix that keeps `JSValue` safe and still avoids memory allocation for argument passing: https://github.com/jasongin/napi-dotnet/commit/f56929f8437918fc14620a6788de5b212b61fe21

Runtime/JSValue.cs
jasongin: I agree memory safety is important.

I think I was confusing the `napi_env` and `napi_handle_scope` in this change (because they're sort of combined now in `JSValueScope`) and assuming a closed scope would result in a `napi_invalid_arg` status from any calls using handles from that scope. But the `napi_env` is still valid, and if the `napi_value` is literally just a pointer, then it could be dereferenced after its scope was closed and memory was freed.

I'll try to find a better solution.

Also we'll need some tests for handle scopes including negative tests. I see there are [some node-addon-api handlescope tests](https://github.com/nodejs/node-addon-api/blob/main/test/handlescope.cc) that could be ported over, though it doesn't look like that includes negative tests.

jasongin:
>Please see the fix that keeps JSValue safe and still avoids memory allocation for argument passing: https://github.com/jasongin/napi-dotnet/commit/f56929f8437918fc14620a6788de5b212b61fe21

Oh thanks! That looks good. I'll merge it into this PR.

vmoroz:
Just a small simplification to put more logic in the `JSCallbackArgs`: https://github.com/jasongin/napi-dotnet/commit/2cd3472d95cc37f092551037f65d2b70dd12bf30

vmoroz (approved):
LGTM

PR #29, opened 2023-02-21
Rewrite and expand README, add TS reference

jasongin:
I thought we'll need a better README, if we're going to start showing this to others soon. So I wrote a summary of all the features we've built so far. A few details are _slightly_ forward-looking, but they'll be accurate eventually.

I'll add details about npm/nuget packages and example projects later, when that's ready.


Docs/typescript.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Object types
+
+| C# Type                     | TypeScript Projection       |
+|-----------------------------|-----------------------------|
+| `class Example`             | `class Example`             |
+| `struct Example`            | `class Example`             |
+| `interface IExample`        | `interface IExample`        |
+| `delegate A Example(B)`     | `example: (B) => A`         |
+| `enum Example { None = 0 }` | `enum Example { None = 0 }` |
```
vmoroz: >enum [](http://example.com/codeflow?start=33&length=4)

enums are generally considered a bad practice in TypeScript and hey recommend not to use them in code because they are no zero abstractions like other type definitions. [https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#enums](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#enums)
I wonder if we should use 'const enum' or unions instead.

README.md
```diff
@@ -1,86 +1,140 @@
...
+```
+```TypeScript
+import { ExampleClass } from './ExampleAssembly';
+ExampleClass.ExampleMethod(...args); // This call is type-checked!
+```
 
-   // Export additional classes from the module by declaring public properties of type `Type`.
-   public Type Another => typeof(Another);
-}
+For reference, there is a paging [listing C# type projections to TypeScript](/Docs/typescript.md).
```
vmoroz: >paging [](http://example.com/codeflow?start=26&length=6)

typo?

Docs/typescript.md
jasongin: > enums are generally considered a bad practice in TypeScript

Do you have a reference for that? I don't see anything in the TS handbook that says they are a bad practice.

I am familiar with the differences between `enum` and `const enum` in TypeScript, and in my experience both are useful in different situations.

For projecting C# enums to JS I [used](https://github.com/jasongin/napi-dotnet/pull/21) non-const enums for two reasons:
  1. Const enums simply do not exist in plain JavaScript. JS code would have to use the numeric constant, which wouldn't be readable or maintainable. Or _maybe_ they could string names but that would have a perf cost and seems unnatural when C# enums are numeric.
  2. It will be useful for JS code to also be able to get the string name of an enum value when needed (e.g. for errors and diagnostics), and that can be done using the reverse mapping provided by TS style enums.

README.md
```diff
@@ -1,86 +1,140 @@
...
+  - C# collections like `IList<T>` and JS collections like `Map<T>` are also passed by reference; access to collection elements is proxied to whichever side the real instance of the collection is on.
+  - JS `TypedArray`s are mapped to C# `Memory<T>` and passed by reference using shared memory (no proxying is needed).
+  - Other types like enums, dates, and delegates are automatically marshaled as one would expect.
+  - Custom marshaling and marshaling hints [may be supported later](https://github.com/jasongin/napi-dotnet/pull/25).
+
+### Optional .NET native AOT compilation
+This library supports hosting the .NET Runtime in the same process as the JavaScript engine. Alternatively, it also supports building [native ahead-of-time (AOT) compiled C#](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/) libraries that are loadable as a JavaScript module _without depending on the .NET Runtime_.
+
+There are advantages and disadvantages to either approach:
+|                 | .NET Runtime | .NET Native AoT |
```
vmoroz: >AoT [](http://example.com/codeflow?start=47&length=3)

typo?

Docs/typescript.md
vmoroz: The link that I posed above points to text that says: "Because of this, it’s a feature which you should know exists, but maybe hold off on using unless you are sure".
In Office code developers started wit using enums, but had to move to unions. I wonder if we must offer an alternative mapping for enums to developers who want to avoid the TS enum overhead.
If "TS enum" require runtime, then do we generate .js file along with the .d.ts for C# code?

README.md
```diff
@@ -1,86 +1,140 @@
...
+ - Automatic marshalling uses shared memory or proxies when possible to minimize the amount of data transferred across the boundary.
+ - Simple calls between JS and C# require **_almost_** zero memory allocation. (Maybe it will be zero eventually.)
+ - Most JavaScript values are represented in C# as small structs (basically containing just a handle to the JS value), which helps avoid memory allocation.
+ - Marshaling code uses modern C# performance features like `Span<T>` and `stackalloc` to minimize heap allocations and copying.
+
+Thanks to these design choices, JS to .NET calls are [more than twice as fast](https://github.com/jasongin/napi-dotnet/pull/23) when compared to `edge-js` using [that project's benchmark](https://github.com/tjanczuk/edge/wiki/Performance).
+
+## Requirements
+ - .NET 6 or later
+    - .NET 7 or later is required for AoT support.
```
vmoroz: >AoT [](http://example.com/codeflow?start=38&length=3)

It maybe not a typo, but should we use consistent spelling of this term. Microsoft docs use 'AOT' spelling: [https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/)

README.md
```diff
@@ -1,86 +1,140 @@
...
+ - Node.js v16 or later
+    - Other JS engines may be supported in the future.
+ - OS: Windows, Mac, or Linux
+    - It should work any platform where .NET 6 is supported.
```
vmoroz: >work any [](http://example.com/codeflow?start=16&length=8)

missing 'on'?

vmoroz: >.NET 6 [](http://example.com/codeflow?start=40&length=6)

Should we add CI tests for it?

README-DEV.md
```diff
@@ -0,0 +1,24 @@
...
+```
+
+Or to run a subset of test cases that match a filter:
+```bash
+dotnet test --filter "DisplayName~hello"
+```
+
+The list of test cases is automatically derived from the set of `.js` files under the `Test/TestCases` directory. Within each subdirectory there, all `.cs` files are compiled into one assembly, then all `.js` test files execute against the assembly.
+
+Most test cases run twice, once for "hosted" CLR mode and once for "aot" ahead-of-time compiled mode with no CLR.
```
vmoroz: >aot [](http://example.com/codeflow?start=68&length=3)

AOT?

Docs/typescript.md
jasongin: > maybe hold off on using unless you are sure

I am sure that I want plain JavaScript code to be able to reference C# enums by name.

> If "TS enum" require runtime, then do we generate .js file along with the .d.ts for C# code?

It's not needed because the runtime enum object is exported by the C# module:

```JavaScript
const { ExampleClass, ExampleEnum } = require('node-api-dotnet')
    .loadAssembly('path/to/ExampleAssembly.dll')
ExampleClass.MethodThatTakesAnEnumParam(ExampleEnum.One);
const result = ExampleClass.PropertyThatReturnsAnEnum;
console.log(`Result: ${ExampleEnum[result]}`);
```

vmoroz (approved):
:shipit:

README.md
jasongin: We should, because many people will be using .NET 6 for a while since it is LTS.

README.md
jasongin: Yes. I was writing "AoT" because o = "of", then I saw .NET uses "AOT" so I switched, but forgot to update some places.

Docs/typescript.md
jasongin: Note the TS handbook describes ["pitfalls" of const enums](https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls) also.

In general const enums are great for internal use in a large project that is all TypeScript only. But for publishing enum values to external consumers, especially when they are JS, non-const enums work better.

README.md
jasongin: Added to the backlog: https://github.com/users/jasongin/projects/1/views/1?pane=issue&itemId=21077422

Docs/typescript.md
jasongin: I added clarification in the doc that they are non-const enums.

I'm going to go ahead and check in this way, since this is documenting the way it is already implemented. We'll update the doc if we ever decide to change how C# enums are handled.

PR #30, opened 2023-02-22
Implement JSError

vmoroz:
Added new type `JSError`. Its functionality was mostly taken from the `Error` type in [node-addon-api](https://github.com/nodejs/node-addon-api/blob/b16c762a190ced903045aa15fbb910eee5c04a97/napi.h#L1732):
- It wraps `JSReference`
- It can store JavaScript `Error` object instance or any other type. Since `napi_ref` can keep only objects and symbols, other types are stores as a special property inside of a wrapper object.

`JSException` wraps `JSError` and can pass it around. All existing code that can throw C# exceptions is converted to use try/catch and propagate the `JSError` via `JSException`, or to throw new JS 'Error' based on C# exception `Message`.

`JSPromise` is changed to use `JSError` to resolve related TODOs.

`JSValueScope` is fixed for Handles to make sure that it uses non-null `napi_env`.


Runtime/JSError.cs
```diff
@@ -0,0 +1,302 @@
...
+    private string? _message;
+    private readonly JSReference? _errorRef;
+
+    private const string ErrorWrapValue = "4bda9e7e-4913-4dbc-95de-891cbf66598e-errorVal";
+    private const string DefaultMessage = "Error in native callback";
+
+    public unsafe JSError(string? message = null, JSErrorType errorType = JSErrorType.Error)
+    {
+        // Fatal error instead of exception to avoid stack overflows
+        using var fataScope = new FatalIfFailedScope();
```
jasongin: Typo: `fatalScope`

Runtime/Hosting/ManagedHost.cs
```diff
@@ -60,6 +60,7 @@ public static napi_value InitializeModule(napi_env env, napi_value exports)
         catch (Exception ex)
         {
             Console.Error.WriteLine($"Failed to load CLR managed host module: {ex}");
```
jasongin: The `Console.Error.WriteLine()` can be changed to `Trace()` now that the error is thrown.

And same in NativeHost.cs. The generated code in ModuleGenerator.cs doesn't have any tracing, but maybe it's OK to just remove the console output from there.

Runtime/JSError.cs
```diff
@@ -0,0 +1,302 @@
...
+namespace NodeApi;
+
+public enum JSErrorType { Error, TypeError, RangeError, SyntaxError }
+
+file record struct JSErrorInfo(string? Message, napi_status Status)
+{
+    public static unsafe JSErrorInfo GetLastErrorInfo()
+    {
+        napi_get_last_error_info((napi_env)JSValueScope.Current,
+                                 out napi_extended_error_info* errorInfo).ThrowIfFailed();
```
jasongin, line 19: While I was working yesterday I hit a bug that caused a stack overflow due to `napi_get_last_error_info()` throwing an exception, then when the exception was caught it would try to get last error info again... I didn't fix it at the time since I expected you were refactoring anyway.

In general, we need to be very careful about throwing new exceptions from within error-handling code.

vmoroz: I agree. In this code we have new `JSError.FatalIfFailedScope` that converts all such exceptions to fatal errors.

Runtime/JSError.cs
```diff
@@ -0,0 +1,302 @@
...
+    {
+        // Fatal error instead of exception to avoid stack overflows
+        using var fataScope = new FatalIfFailedScope();
+
+        // We must retrieve the last error info before doing anything else because
+        // doing anything else will replace the last error info.
+        JSErrorInfo errorInfo = JSErrorInfo.GetLastErrorInfo();
+
+        // A pending JS exception takes precedence over any internal error status.
+        if (IsExceptionPending())
```
jasongin, line 54: `IsExceptionPending()` may throw a new exception if it fails. Either it shouldn't do that, or this code should call the lower layer API directly and handle the status result, or at least catch the exception.

Runtime/JSError.cs
vmoroz: Implementing the scope was easier than going after all interop functions and have their alternative handling,

Runtime/JSError.cs
jasongin: Oh, now that I've reviewed as far as `FatalIfFailedScope` I guess that resolves these issues.

vmoroz: Such exceptions will result in fatal error in this code. It is aligned with `node-addon-api`. They use `NAPI_FATAL_IF_FAILED` macro is such cases.

vmoroz: E.g. see: https://github.com/nodejs/node-addon-api/blob/main/napi-inl.h#L2670

Runtime/JSError.cs
jasongin: Never mind, mitigated with `FatalIfFailedScope`.

Runtime/JSError.cs
```diff
@@ -0,0 +1,302 @@
...
+
+public struct JSError
+{
+    private string? _message;
+    private readonly JSReference? _errorRef;
+
+    private const string ErrorWrapValue = "4bda9e7e-4913-4dbc-95de-891cbf66598e-errorVal";
+    private const string DefaultMessage = "Error in native callback";
+
+    public unsafe JSError(string? message = null, JSErrorType errorType = JSErrorType.Error)
```
jasongin: Should it have an optional `code` parameter?

jasongin:
Maybe not in this PR, but eventually it would be great to propagate stack traces across .NET / JS.

Runtime/JSError.cs
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+
+    public JSValue? Value
+    {
+        get
+        {
+            JSValue? error = _errorRef?.GetValue();
+            if (error is null)
+                return null;
```
jasongin: I think it would be more convenient & consistent if the `JSError.Value` property is not nullable, then this could return undefined. The `Message` property returns a default non-null value.

Runtime/JSNativeApi.cs
```diff
@@ -833,45 +731,68 @@ public static void Seal(this JSValue thisValue)
...
+            JSCallbackArgs.GetDataAndLength(env, callbackInfo, out object? data, out int length);
+            Span<napi_value> args = stackalloc napi_value[length];
+            JSPropertyDescriptor desc = (JSPropertyDescriptor)data!;
+            return (napi_value)desc.Method!.Invoke(
+                new JSCallbackArgs(scope, callbackInfo, args, desc.Data));
+        }
+        catch (Exception ex)
+        {
+            JSError.ThrowError(ex);
+            return napi_value.Null;
```
jasongin, line 756: I think undefined would be more correct than null... though since the error was thrown the caller never gets the result anyway.

Runtime/JSValue.cs
```diff
@@ -342,3 +346,9 @@ public override int GetHashCode()
             "Hashing JS values is not supported. Use JSSet or JSMap instead.");
     }
 }
+
+public static class JSValueExtensions
+{
+    public static napi_value AsNapiValueOrNull(this JSValue? value)
```
jasongin, line 352: Can this be internal?

Runtime/JSValue.cs
```diff
@@ -156,17 +156,21 @@ public static JSValue CreateFunction(string name, JSCallback callback)
...
-    public static JSValue CreateRangeError(JSValue code, JSValue message)
-        => napi_create_range_error(Env, (napi_value)code, (napi_value)message, out napi_value result).ThrowIfFailed(result);
+    public static JSValue CreateRangeError(JSValue? code, JSValue message)
+        => napi_create_range_error(Env, code.AsNapiValueOrNull(), (napi_value)message,
+            out napi_value result).ThrowIfFailed(result);
 
-    public static JSValue CreateSyntaxError(JSValue code, JSValue message)
-        => node_api_create_syntax_error(Env, (napi_value)code, (napi_value)message, out napi_value result).ThrowIfFailed(result);
+    public static JSValue CreateSyntaxError(JSValue? code, JSValue message)
+        => node_api_create_syntax_error(Env, code.AsNapiValueOrNull(), (napi_value)message,
```
jasongin, line 172: Do these APIs not accept `undefined` as a valid value for the code?

In JS, `new Error().code === undefined`.

jasongin (approved):

Runtime/JSValue.cs
vmoroz: In Node.JS they expect either `null` or a `string`: https://github.com/nodejs/node/blob/655b070d31efeea2f5580ca221167f8abc43f0e8/src/js_native_api_v8.cc#L1583

Runtime/JSValue.cs
vmoroz: Yes, I will change it to internal.

Runtime/JSNativeApi.cs
vmoroz: In Node.JS they expect callbacks to return `NULL` if no data should be returned. The `undefined` will require us to call extra Node-APu function. If we return a non-null value, then the Node.JS code will set result value: https://github.com/nodejs/node/blob/655b070d31efeea2f5580ca221167f8abc43f0e8/src/js_native_api_v8.cc#L323-L325

Runtime/JSError.cs
vmoroz: Good point! Let me change it.

vmoroz:
> Maybe not in this PR, but eventually it would be great to propagate stack traces across .NET / JS.

I will see how to do it in a follow up PR.

PR #31, opened 2023-02-22
Move test common/index.js one level up

vmoroz:
It is a simple mechanical change to move common/index.js used for testing one directory up.
The goal is to make this file common for all tests and not only for the `node-addon-api` where it is started.


jasongin (approved):
> .../TestCases/node-addon-api/common/index.js → Test/TestCases/common/index.js

Make sure the `TestBuilder` does not detect this as a test case now. You may need to add code to exclude it as a special case.

vmoroz:
> > .../TestCases/node-addon-api/common/index.js → Test/TestCases/common/index.js
> 
> Make sure the `TestBuilder` does not detect this as a test case now. You may need to add code to exclude it as a special case.

Sorry, I did not notice your comment. I will fix it in the future PR. 
It was previously also detected as a test case. Since it always passes, there were no extra action needed.

PR #32, opened 2023-02-27
Use expressions for both static and dynamic binding

jasongin:
Following are the major new or significantly updated components in this PR:
 - The `JSMarshaler` class is the core of this change. It builds expression trees for adapting property/method access and converting values between .NET and JS. It is used both during compile-time source generation (for static binding / AOT) and runtime dynamic binding.
 - The `SymbolExtensions` class converts compile-time source-generator symbols like `ITypeSymbol` and `IMethodSymbol` to their equivalent runtime reflection types (`Type`, `MethodInfo`) by creating "symbolic" types without any method bodies. This enables expression trees to be built at compile-time, even though they require the runtime reflection types.
 - The `ExpressionExtensions` class generates C# code from expression trees. The code is a little ugly but it's pretty straightforward and easy to update if needed for new types of expressions.
 - The `ModuleGenerator` class is updated to use the above classes to generate most of the C# code necessary for exporting JS at build time. (A little bit of C# code like basic module registration is still generated as text. The previous `AdapterGenerator` helper class that did most of the C# generation is deleted.)
 - The `JSInterface` class is an abstract base class for JS adapters that implement .NET interfacers. (Previously this class was generated by the source generator.)
    - It requires some use of Reflection.Emit because it's not possible to provide a delegate directly as a method implementation for a dynamically built class.
    - It's not fully implemented, but I plan to have the emitted IL just invoke a delegate created from expressions.
 - The `AssemblyExporter` class handles dynamically loading assemblies and invoking .NET APIs using adapters generated by a `JSMarshaler`.
   - This class is far from complete, as indicated by several "TODOs". There is a lot more work to be done for dynamic invocation. In particular, overload resolution will be important. I have some ideas about how to do that, but this change was getting big enough already.


vmoroz (approved):
:shipit:

PR #33, opened 2023-02-28
Move TS generation to command-line tool

jasongin:
I'm moving the generation of TypeScript type definitions away from being a C# compile-time source generator to a standalone executable command. Two reasons:

1. C# source generators are not supposed to generate other files (or do any I/O at all). They are only meant to add code to the current compilation context.
   - While working on other things, I noticed VS was getting extremely sluggish. Disabling the type definitions part of the source-generator resolved the issue. That is consistent with the documented guidance about source generators: writing other files from the generator can cause perf problems. That motivated me to make this change sooner rather than later.
   
2. Now that this project is starting to support [invoking pre-built .NET APIs](https://github.com/jasongin/napi-dotnet/pull/32), it will be helpful to generate type definitions for those as well. That means the generator can't run during the initial compilation; it needs to reflect over a built assembly.

The type-definitions generator code will continue to live in the same `NodeApi.Generator` assembly. It just doesn't run in the same process as the C# compiler anymore. The assembly is now built as an EXE, though the source-generator part of it is still used as a class library by the C# compiler.

This change required updating the type definitions generator code to work on reflection types (`Type`, `MethodInfo`, ...) rather than compile-time symbols (`ITypeSymbol`, `IMethodSymbol`, ...).

For now the TS generator command-line interface is super simple. It just takes two arguments: the input assembly `.dll` path and output `.d.ts` path. We can update it later to support a richer command-line interface with options.

Also we will need MSBuild tasks/targets to automate integrating the generator into a project.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -9,6 +9,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.3" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.4.0" />
     <PackageVersion Include="Nerdbank.GitVersioning" Version="3.5.119" />
+    <PackageVersion Include="System.Reflection.MetadataLoadContext" Version="7.0.0" />
     <PackageVersion Include="xunit" Version="2.4.2" />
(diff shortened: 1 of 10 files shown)


vmoroz (approved):
:shipit:

PR #34, opened 2023-03-01
Dynamic binding to .NET APIs with overload resolution

jasongin:
This change fills out more of the implementation of dynamic binding to .NET assemblies, so that it supports most kinds of APIs including classes, structs, enums, properties, and methods. (Dynamic binding for .NET interfaces is still not fully implemented.)

And it adds basic overload resolution for constructors and methods. So for example now you can do this:
```JavaScript
const version = new dotnet.Version(1, 2, 3); // Invoke overloaded constructor with args.
assert.strictEqual(version.ToString(), '1.2.3');
assert.strictEqual(version + '.4', '1.2.3.4'); // Implicit call to .NET ToString()
```

Overload resolution looks first at the number of supplied arguments, then if the call is still ambiguous it tries to match the argument JS value types (number, string, object, etc.) to the .NET parameter types. It does not yet do any matching on object _class_; that will probably be necessary eventually.

One lower level change worth noting is how the constructor callback delegate is now also a `JSCallback`. (A `Func<T>` may still be provided to the class builder, though it gets immediately wrapped in the general callback delegate type.) Overload resolution would have been very messy if it had to work with many types of callback delegates, so that change kept it much simpler. Related to that, constructors and standalone functions now support an optional data object via `JSCallbackDescriptor`. The overload resolution table is passed to the callback that way, but it can be used for other things too.

I haven't done broad testing against many kinds of .NET APIs yet, but I expect there will be a lot of problems and limitations that we'll have to triage and work through to make this more seamless. Another issue now is that binding is done on a per-type basis, not per-member, so if any one property or method fails to bind then the entire type fails to load. That can be fixed later with some proxy magic.

### DIFF
--- a/Runtime/JSNativeApi.cs
+++ b/Runtime/JSNativeApi.cs
@@ -362,12 +362,14 @@ public static partial class JSNativeApi
         }
     }
 
-    public static unsafe JSValue DefineClass(ReadOnlySpan<byte> utf8Name, JSCallback callback, params JSPropertyDescriptor[] descriptors)
+    public static unsafe JSValue DefineClass(
+        ReadOnlySpan<byte> utf8Name,
+        JSCallbackDescriptor constructorDescriptor,
+        params JSPropertyDescriptor[] propertyDescriptors)
     {
-        int nameLength = utf8Name.Length;
-        GCHandle callbackHandle = GCHandle.Alloc(callback);
+        GCHandle callbackHandle = GCHandle.Alloc(constructorDescriptor);
         JSValue? func = null;
-        nint[] handles = ToUnmanagedPropertyDescriptors(utf8Name, descriptors, (name, count, descriptorsPtr) =>
+        nint[] handles = ToUnmanagedPropertyDescriptors(utf8Name, propertyDescriptors, (name, count, descriptorsPtr) =>
         {
             func = DefineClass(name, new napi_callback(&InvokeJSCallback), (nint)callbackHandle, count, descriptorsPtr);
         });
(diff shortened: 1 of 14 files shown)


Runtime/JSNativeApi.cs
```diff
@@ -362,12 +362,14 @@ public static bool InstanceOf(this JSValue thisValue, JSValue constructor)
...
 
-    public static unsafe JSValue DefineClass(ReadOnlySpan<byte> utf8Name, JSCallback callback, params JSPropertyDescriptor[] descriptors)
+    public static unsafe JSValue DefineClass(
+        ReadOnlySpan<byte> utf8Name,
+        JSCallbackDescriptor constructorDescriptor,
+        params JSPropertyDescriptor[] propertyDescriptors)
     {
-        int nameLength = utf8Name.Length;
-        GCHandle callbackHandle = GCHandle.Alloc(callback);
+        GCHandle callbackHandle = GCHandle.Alloc(constructorDescriptor);
```
vmoroz, line 370: >GCHandle.Alloc(constructorDescriptor); [](http://example.com/codeflow?start=34&length=38)

How does it work for structs?

vmoroz: I see that we do the same for the `JSPropertyDescriptor`. I guess it is being boxed.

vmoroz (approved):
:shipit:

Runtime/JSNativeApi.cs
jasongin: Yes, it gets boxed since `GCHandle.Alloc()` takes an `object`. Then it's unboxed by casting the callback args `object data` value to `JSCallbackDescriptor`.

`JSPropertyDescriptor` is not quite the same because it gets copied into an array of `napi_property_descriptor`.

PR #36, opened 2023-03-04
Layering proposal

jasongin:
This is a proposal for assembly/namespace naming and layering. It is not vastly different from what we have now, but includes some additional namespaces for organization and more explicitly defines the layering strategy.

Also after more consideration I want to try to move the .NET hosting / dynamic-binding code into a separate assembly, so that the layering is enforced by the compiler and it is clearer what code gets AOT-compiled and what does not. (I know I said the opposite earlier today. I still reserve the right to take back that suggestion if it turns out to cause too much trouble with AOT building/loading or something like that.)


vmoroz:
It looks like a good start! There are a few things that I would suggest doing differently:
- It seems strange that `JSObject` and `JSArray` are in different namespaces. They are both helpers for the basic JavaScript support. It would be better to keep them both in the same core namespace.
- Then the next question is if it is worth having the `.Collections` namespace at all.
- Should we rather have a namespace for JS built-in objects?
- `JSContext` is currently referenced from `JSReference`. It is strange that they are in a different namespace. Should we split the `JSContext` into two classes? One for the core purposes, and another for more advanced interop?
- I wonder if we can also skip introducing the `.NodeJS` namespace. I hope that we can achieve this differentiation in the API loader instead.
- The `JSClassBuilder` and the `JSAsyncScope` should be part of the core namespace.

What if we can consider the following criteria for the namespace and assembly separation:
- Use in Node.JS vs another environment like React Native.
- Core JS APIs/object vs interop with C# data types.
- Runtime vs Compile time utilities.

jasongin:
> It seems strange that `JSObject` and `JSArray` are in different namespaces. They are both helpers for the basic JavaScript support. It would be better to keep them both in the same core namespace.

Yes I debated that with myself. But it would also be strange if there is a collections namespace while `JSArray` is not in it.

> Then the next question is if it is worth having the `.Collections` namespace at all.

Perhaps not; I don't have a strong opinion about it. But .NET sets the convention of having collections in a namespace. And I was erring on the side of not leaving too much stuff at the root and collections seemed like an obvious grouping.

> - Should we rather have a namespace for JS built-in objects?

I'm not sure what you mean by "built-in". Would that include classes like Set, Map, Date, Proxy? They are part of the JS standard. Or do you just mean the basic value-types? The latter would not include arrays since those are technically object value-type.


> - `JSContext` is currently referenced from `JSReference`. It is strange that they are in a different namespace.

There are many inter-dependencies within the main library assembly. Trying to enforce layering at the namespace level would lead to nearly everything being in the same namespace, and would quickly be invalidated by future code changes since the compiler doesn't enforce it.

> Should we split the `JSContext` into two classes? One for the core purposes, and another for more advanced interop?

Possibly. I realized recently that I'm going to need to do some refactoring there anyway to support multiple .NET "module contexts" within what Node.js treats as a single add-on.

> - I wonder if we can also skip introducing the `.NodeJS` namespace. I hope that we can achieve this differentiation in the API loader instead.

I don't understand. But that namespace subdivision is not important.

> - The `JSClassBuilder` and the `JSAsyncScope` should be part of the core namespace.

I don't agree. Then `JSStructBuilder` and similar and related types would logically follow it, then at that point nearly everything is back in the root namespace. My intention was to move out all the types that most developers won't need when they are just using `[JSExport]` generated module/class registration or doing basic read/write interactions with JS values.

> What if we can consider the following criteria for the namespace and assembly separation:
> 
> - Use in Node.JS vs another environment like React Native.

It's hard for me to predict the requirements for that environment until we have an example.

> - Core JS APIs/object vs interop with C# data types.
> 
> - Runtime vs Compile time utilities.

Those two points summarize what I'm attempting here, aside from the collections and NodeJS subdivisions which are less important.

vmoroz:
> > * Should we rather have a namespace for JS built-in objects?
> 
> I'm not sure what you mean by "built-in". Would that include classes like Set, Map, Date, Proxy? They are part of the JS standard. Or do you just mean the basic value-types? The latter would not include arrays since those are technically object value-type.

You are right, it seems to be a rather artificial split in that case. It makes sense to keep them all as a part of the core JS support code. 

> 
> > * `JSContext` is currently referenced from `JSReference`. It is strange that they are in a different namespace.
> 
> There are many inter-dependencies within the main library assembly. Trying to enforce layering at the namespace level would lead to nearly everything being in the same namespace, and would quickly be invalidated by future code changes since the compiler doesn't enforce it.

Good point!

> > * The `JSClassBuilder` and the `JSAsyncScope` should be part of the core namespace.
> 
> I don't agree. Then `JSStructBuilder` and similar and related types would logically follow it, then at that point nearly everything is back in the root namespace. My intention was to move out all the types that most developers won't need when they are just using `[JSExport]` generated module/class registration or doing basic read/write interactions with JS values.
> 
Good point!

jasongin:
It sounds like we're mostly in sync then, aside from maybe some minor details.

Another consideration is that the source code directory structure does not have to exactly match the namespace structure -- e.g. there could be a `Collections` subfolder without making it a sub-namespace.

I do think we should revisit this code organization when we get into embedding JS inside a C# application, to make sure everything still makes sense in that environment.

I'd like to move forward with this fairly soon, so we can get all the naming set up, and because I have some more changes in the pipeline (nuget/npm packaging) that will depend on the assembly split between core library and .NET hosting.

Note in the [PR that implements this proposal](https://github.com/jasongin/napi-dotnet/pull/37) I did not change any namespaces yet, so it iss not blocked on deciding whether or not to have Collections or NodeJS namespace subdivisions.

vmoroz (approved):

jasongin:
Closing because this is implemented by https://github.com/jasongin/napi-dotnet/pull/37. That includes some updates to the namespace organization after discussion.

PR #37, opened 2023-03-04
Organize & rename projects and assemblies

jasongin:
Rename and reorganize projects and assemblies according to the planned names and layering proposal in https://github.com/jasongin/napi-dotnet/pull/36

Namespaces are not updated in this PR, as that will be easier to review separately. So most source files here are merely moved, not modified.

There are a few source code changes aside from the obvious `.csproj`/`.props` changes:
 - Update test build orchestration code to deal with the separation of the `DotNetHost` assembly, along with some project/assembly names and paths referenced in that code.
 - Remove code analysis suppressions in hosting code that are no longer needed now that that code is not in an assembly that is AOT-compiled.
 - Refactor the dynamic emit code in `JSInterface` out to `JSInterfaceMarshaler`. The former must live in the main assembly while the latter belongs in the `DotNetHost` assembly.
 - Fix the assembly resolve handler in `ManagedHost` to also support the new assembly (and use the more modern API).
 - Fix the error-handling code in `NatieHost`; the high-level `ThrowError` API didn't work without any callback scope, so it is switched to the low-level API.


vmoroz:
It seems that the DotNetHost code is still part of the runtime. Should it be moved?

Like I mentioned in #36, separation of all collections into a separate folder looks rather artificial. 
It would be better to keep core types such as `JSArray`, `JSMap`, etc. along with other JS core types, while splitting up the C# type interop into a separate folder. Even if it means that we must come up with new type names. E.g. instead of `JSArray` use term `JSArrayLike`. I think this term is often used in JS for types that behave like arrays, but are not true JS arrays.

The rest looks awesome!

jasongin:
> It seems that the DotNetHost code is still part of the runtime. Should it be moved?

`ManagedHost` and supporting classes are moved into a new `NodeApi.DotNetHost` assembly. `NativeHost` must stay in the main assembly because it gets AOT-compiled with the other library code. Separating those may be technically possible but I think not worth having a whole assembly with just `NativeHost` + `HostFxr` types. (Anyway `NativeHost` is going to get about 50% smaller in my next PR; I'll explain then.)

> It would be better to keep core types such as JSArray, JSMap, etc. along with other JS core types, while splitting up the C# type interop into a separate folder.

It is possible to change the C# type interop methods like `JSArray.AsList<T>()` to extension methods. So there would be `JSArrayExtensions` and similar extension classes in the interop namespace. And the interop code in `.Proxy.cs` is all static internal, so it can go anywhere.

vmoroz (approved):

jasongin:
I eliminated the `Collections` subfolder, refactored the collection proxies as discussed above, and updated the diagram:

PR #38, opened 2023-03-06
Update namespaces

jasongin:
- Prepend `Microsoft.JavaScript.` to all `NodeApi` namespaces.
 - Add `Interop` namespace for that subfolder in main library.
 - Rename `Hosting` namespace to `DotNetHost`.

I am NOT creating a `Microsoft.JavaScript.NodeApi.Native` namespace yet, because that will require more refactoring. Lots of code depends on extension methods in `JSNativeApi`. I think those should be refactored to methods or extension methods in the root namespace, then `JSNativeApi` should be internal. Also maybe code within `JSNativeApi` can be organized into more separate types, since they will all be in an internal namespace.

### DIFF
--- a/src/NodeApi.DotNetHost/AssemblyExporter.cs
+++ b/src/NodeApi.DotNetHost/AssemblyExporter.cs
@@ -3,10 +3,11 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using Microsoft.JavaScript.NodeApi.Interop;
 
-using static NodeApi.Hosting.ManagedHost;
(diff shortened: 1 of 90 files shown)


vmoroz:
I am currently changing the way we do the interop with Node-API functions. Could you postpone changes to `JSNativeApi` for a couple of days? I agree with you: we should probably delete this class completely and spread its code between other types.

vmoroz (approved):
:shipit:

jasongin:
> Could you postpone changes to JSNativeApi for a couple of days?

I made some minor changes in that area over the weekend, for this PR: https://github.com/jasongin/napi-dotnet/pull/39

After that I don't expect to be working there any time soon.

PR #39, opened 2023-03-06
NuGet and npm packaging

jasongin:
**I suggest reviewing the 3 new example projects first.** They are intended to provide minimal canonical examples of how to consume the nuget and npm packages, for the 3 main JS->.NET scenarios supported so far:
  - `examples/aot-module`
  - `examples/dotnet-module`
  - `examples/dotnet-dynamic`

Notes about the rest of the changes:
 - Enable automatic build versions based on git commit (NerdBank.GitVersioning).
 - Set up 3 nuget packages:
    - `Microsoft.JavaScript.NodeApi`
    - `Microsoft.JavaScript.NodeApi.DotNetHost`
    - `Microsoft.JavaScript.NodeApi.Generator`
    - Add `.props` and `.targets` files to nuget packages to streamline integration for projects that consume the packages.
 - Set up npm package: `node-api-dotnet`
    - This is an unusually structured npm package, so a normal `.npmrc` file wouldn't be sufficient. A custom `pack.js` script stages all the required script and binary files before calling `npm pack`.
    - By default it packs binaries only for the current system RuntimeIdentifier, but it also supports a list of rids for creation of a multi-platform package.
    - The `pack.js` is called from MSBuild so both npm and nuget packages can be built with a single `dotnet pack` command.
  - Change the Node API `NativeHost` -> `ManagedHost` activation flow so that it gets the path to the managed host assembly from JS code (`index.js` bundled in the npm package) instead of inferring the location.
    - This will enable the npm package to support multiple target framework versions, though only `net7.0` is built now.
    - This greatly reduces the amount of messy P/Invoke code required in `NativeHost`. Previously ~100 lines of code were required just to find the path to itself, so that it could infer the path to the adjacent managed host assembly.
    - Exporting a JS callback method from `NativeHost` exposed a pre-existing problem with the way both `NativeHost` and `ManagedHost` interacted with Node API module context. I had to create "NoContext" variations of the `JSNativeApi` callback adapters to support native host callbacks that don't rely on a Node API module context. (But there is still a problem for the managed host loading multiple modules; I'm not solving that now.)


vmoroz:
>             NativeLibrary.SetDllImportResolver(

Since this call is per process, we cannot really do it in user modules. I am changing the code to avoid this call.

---
Refers to: src/NodeApi/Native/JSNativeApi.Interop.cs:26 in 8c79fac. [](commit_id = 8c79fac86cffe7db17d42b632df8e29ad673916f, deletion_comment = False)

src/NodeApi/DotNetHost/NativeHost.cs
```diff
@@ -43,28 +40,19 @@ public static napi_value InitializeModule(napi_env env, napi_value exports)
...
-            try
-            {
-                exports = host.InitializeManagedHost(env, exports);
-            }
-            catch
-            {
-                host.Dispose();
-                throw;
-            }
+            using JSValueScope scope = new(JSValueScopeType.RootNoContext, env);
```
vmoroz: > using JSValueScope scop [](http://example.com/codeflow?start=11&length=24)

Why it was moved inside of try block? Does it mean that the code in the catch block will now always crash the app instead of reporting errors? The `JSValue.CreateError` requires a `JSValueScope`.

src/NodeApi/DotNetHost/NativeHost.cs
```diff
@@ -43,28 +40,19 @@ public static napi_value InitializeModule(napi_env env, napi_value exports)
...
+            JSNativeApi.Interop.Initialize();
+
+            NativeHost host = new();
 
+            // Do not use JSModuleBuilder here because it relies on having a current context.
+            // But the context will be set by the managed host.
+            new JSValue(exports, scope).DefineProperties(
+                // The package index.js will invoke the initialize method with the path to
+                // the manged host assembly.
```
vmoroz: >manged [](http://example.com/codeflow?start=23&length=6)

typo?

examples/dotnet-module/README.md
```diff
@@ -0,0 +1,10 @@
+
+## Minimal Example .NET Node Module
+The `Example.cs` class defines a Node.js add-on module that runs on .NET. The `example.js` script loads that .NET module and calls a method on it. The script has access to type definitions and doc-comments for the module's APIs via the auto-generated `.d.ts` file.
```
vmoroz: >The [](http://example.com/codeflow?start=0&length=3)

nit: can we avoid long text lines? MD format allows using new lines in the same paragraph.

vmoroz (approved):
:shipit:

src/NodeApi/DotNetHost/NativeHost.cs
jasongin: You're right. I was moving things around here to troubleshoot the module context issue, but the scope definitely should be outside the try block. Fixed.

examples/dotnet-module/README.md
jasongin: Fixed in all the MD files.

PR #40, opened 2023-03-08
JSNativeApi.Interop to bind functions from a specified module.

vmoroz:
Currently we bind Node-API C functions from the current running module by calling `SetDllImportResolver`.
It works, but it will not allow us using our library outside of Node.JS:
- The Node-API in React Native is going to be exposed from a specific DLL.
- In standalone C# application it may have its own call to `SetDllImportResolver` and thus we cannot do it in library.

In this PR we do not solve the problem completely, but make initial important steps:
- Implement an alternative binding for Node-API C functions by calling `NativeLibrary.GetExport`.
- Remove Node-API DLL resolver from the `SetDllImportResolver`.

I wonder if we can test perf for the new approach against previous. I do not want it causing regressions.

To complete the change, we need to:
- Move `Inperop.Initialize()` call from `JSContext` to code that creates it.
- Implement different binding for `HostFxr` functions and remove the `SetDllImportResolver` call.

In this PR I also did some formatting changes to fit lines into the 100-character limit for the files I was changing.
Sorry, it unnecessary produced more churn than needed.

Spelling of word 'marshler' was changed to 'marshaller' in the code and docs.

### DIFF
--- a/src/NodeApi/Native/JSNativeApi.Interop.cs
+++ b/src/NodeApi/Native/JSNativeApi.Interop.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.Marshalling;
 using System.Security;
 
 namespace Microsoft.JavaScript.NodeApi;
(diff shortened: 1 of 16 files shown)


src/NodeApi/Native/JSNativeApi.Interop.cs
```diff
@@ -220,451 +386,1398 @@ public struct napi_type_tag
...
+            }
+        }
 
-        [LibraryImport(nameof(NodeApi)), UnmanagedCallConv(CallConvs = new[] { typeof(CallConvCdecl) })]
-        internal static partial napi_status napi_get_version(napi_env env, out uint result);
+        internal static napi_status napi_set_instance_data(
+            napi_env env, nint data, napi_finalize finalize_cb, nint finalize_hint)
+        {
+            nint funcHandle = Current!.GetExport(
+                FunctionId.napi_set_instance_data, nameof(napi_set_instance_data));
```
jasongin: This can rely on CallerMemberName like the others.

src/NodeApi/Native/JSNativeApi.Interop.cs
```diff
@@ -220,451 +386,1398 @@ public struct napi_type_tag
...
+            [CallerMemberName] string functionName = "")
+        {
+            nint funcHandle = interop!.GetExport(functionId, functionName);
+            var funcDelegate = (delegate* unmanaged[Cdecl]<napi_env, nint, napi_status>)funcHandle;
+            return funcDelegate(env, value);
+        }
 
-        [LibraryImport(nameof(NodeApi)), UnmanagedCallConv(CallConvs = new[] { typeof(CallConvCdecl) })]
-        internal static partial napi_status napi_object_seal(napi_env env, napi_value @object);
+        private static napi_status CallInterop(
```
jasongin, line 1664: This method is not marked `unsafe` like the others. I guess none of them need it because the containing class is `unsafe`?

src/NodeApi/Native/JSNativeApi.Interop.cs
```diff
@@ -220,451 +386,1398 @@ public struct napi_type_tag
...
+            {
+                return funcDelegate(
+                    env, value1, value2, value3, value4, value5, (nint)result_native);
+            }
+        }
 
-        [LibraryImport(nameof(NodeApi)), UnmanagedCallConv(CallConvs = new[] { typeof(CallConvCdecl) })]
-        internal static partial napi_status napi_check_object_type_tag(napi_env env, napi_value value,
-           in napi_type_tag type_tag, out c_bool result);
+        private static unsafe napi_status CallInterop<TResult>(
```
jasongin: The choice for whether to create a `CallInterop` overload for a given signature or not appears to be based on whether there is more than one API that shares the signature. Most of the overloads have multiple references, except a few like this one are only referenced once each. Should they be inlined with the callers?

src/NodeApi/Native/JSNativeApi.Interop.cs
```diff
@@ -220,451 +386,1398 @@ public struct napi_type_tag
...
+            napi_env env,
+            string? value1,
+            string? value2,
+            [CallerMemberName] string functionName = "")
+        {
+            nint funcHandle = interop!.GetExport(functionId, functionName);
+            var funcDelegate = (delegate* unmanaged[Cdecl]<
+                napi_env, nint, nint, napi_status>)funcHandle;
+
+            Utf8StringMarshaller.ManagedToUnmanagedIn value1_marshaller = new();
```
jasongin, line 1675: Maybe `JSMarshaler` should be spelled with two `L`s also? I wasn't sure about that spelling.

src/NodeApi/Native/JSNativeApi.Interop.cs
```diff
@@ -11,36 +12,200 @@ public static partial class JSNativeApi
...
+        public static Interop? Current { get; set; }
+
+        public Interop(nint libraryHandle) => _libraryHandle = libraryHandle;
+
+        public static void Initialize(nint libraryHandle)
         {
             if (s_initialized) return;
             s_initialized = true;
 
+            Current = new Interop(libraryHandle);
```
jasongin: Should it be a static class instead of singleton? I can't think of any reason it needs to be an instance.

Static method calls are slightly faster since there is one less level of indirection and no need for a null check.

jasongin (approved):

src/NodeApi/Native/JSNativeApi.Interop.cs
vmoroz: Good point! I fixed all similar places.

src/NodeApi/Native/JSNativeApi.Interop.cs
vmoroz: Yep, removing all the extra `unsafe` did not affect the compilation.

src/NodeApi/Native/JSNativeApi.Interop.cs
vmoroz: I looked through the [marshalling .Net docs](https://learn.microsoft.com/en-us/dotnet/framework/interop/default-marshalling-behavior) and they often mix them, but it seems that the most common pattern is that they use term "Marshal" as a verb with one "L", but they write "Marshalling" and "Marshaller" with two "L". I added the change to this PR.

src/NodeApi/Native/JSNativeApi.Interop.cs
vmoroz: Thank you for the suggested change! I merged it and I also changed from `enum` fields to a `struct` to avoid extra bounds checking.

src/NodeApi/Native/JSNativeApi.Interop.cs
```diff
@@ -1323,31 +1319,31 @@ internal static napi_status napi_detach_arraybuffer(napi_env env, napi_value arr
...
         internal static napi_status napi_object_freeze(napi_env env, napi_value js_object)
-            => CallInterop(FunctionId.napi_object_freeze, env, js_object.Handle);
+            => CallInterop(ref s_fields.napi_object_freeze, env, js_object.Handle);
 
         internal static napi_status napi_object_seal(napi_env env, napi_value js_object)
-            => CallInterop(FunctionId.napi_object_seal, env, js_object.Handle);
+            => CallInterop(ref s_fields.napi_object_seal, env, js_object.Handle);
 
-        private static nint GetExport(FunctionId functionId, [CallerMemberName] string functionName = "")
+        private static nint GetExport(ref nint field, [CallerMemberName] string functionName = "")
```
jasongin, line 1332: Is there a reason to pass the fields by ref when they are just `nint`?

jasongin: Oh I see, because the field may be initialized on first use.

vmoroz: This function also initializes the passed fields. This is why we need it to be an in-out parameter.

src/NodeApi/Native/JSNativeApi.Interop.cs
vmoroz: Good point! I removed this helper method.

vmoroz: My initial implementation was always using helper methods. They were all generic, but to work around issues with using generics in unmanaged delegate I had to start to get rid of some of them.

PR #41, opened 2023-03-09
.NET 6 support

jasongin:
This was more complicated than I expected, mostly due to the way the .NET 6 tests must still depend on the .NET 7 native host, because AOT compilation doesn't work with .NET 6.

 - Use `NativeLibrary.GetExport()` to bind the `HostFxr` function calls. The delegates are not cached because they are only called once each per process, when starting .NET.
 - Use P/Invoke to get the current module handle because `NativeLibrary.GetMainProgramHandle()` is not available in .NET 6.
 - Use alternative string marshaling code because `Utf8StringMarshaller` is not available in .NET 6.
 - Replace `nint.Zero` with `default` because the former is not available in .NET 6.
 - Set `LangVersion=10` to support compiling code with .NET 6 SDK (without AOT support).
   - This was required so that `net6.0` tests can drive MSBuild to build the auto-generated test projects.
   - We were using three C# 11 language features:
      - Static `Unwrap` interface method - replaced with a delegate parameter to the base class constructor.
      - Implicit default initialization of struct fields - easily initialized explicitly.
      - `CallerArgumentName` in tests code - replaced with explicit `nameof()`.
 - Update `pack.js` script for npm package to include `net6.0` target assemblies.
 - Update various parts of the test runner to support running tests against both `net6.0` and `net7.0` targets.
 - Update publish/pack MSBuild properties to support multi-targeting. Unfortunately it's no longer possible to have `dotnet pack` automatically publish first. So the examples all require an additional setup command; I updated the READMEs.
 - Update `build.yml` so the GH actions install both .NET 6 and 7 and run tests against both.

### DIFF
--- a/src/NodeApi/DotNetHost/HostFxr.cs
+++ b/src/NodeApi/DotNetHost/HostFxr.cs
@@ -1,7 +1,6 @@
 
 using System;
 using System.IO;
-using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
 
@@ -143,11 +142,8 @@ internal static partial class HostFxr
         */
     }
 
-    // Note this is CORECLR_DELEGATE_CALLTYPE, which is stdcall on Windows.
-    // See https://github.com/dotnet/runtime/blob/main/src/native/corehost/coreclr_delegates.h
     // The returned function pointer must be converted to a specific delegate via
     // Marshal.GetDelegateForFunctionPointer().
-    [UnmanagedFunctionPointer(CallingConvention.StdCall)]
     public unsafe delegate hostfxr_status load_assembly_and_get_function_pointer(
         byte* assemblyPath, // UTF-16 on Windows, UTF-8 elsewhere
         byte* typeName,     // UTF-16 on Windows, UTF-8 elsewhere
@@ -163,7 +159,7 @@ internal static partial class HostFxr
     {
         nint funcHandle = NativeLibrary.GetExport(
             Handle, nameof(hostfxr_initialize_for_runtime_config));
-        var funcDelegate = (delegate* unmanaged[Cdecl]<
+        var funcDelegate = (delegate* unmanaged[Cdecl]< // HOSTFXR_CALLTYPE = cdecl
                 byte*, hostfxr_initialize_parameters*, hostfxr_handle*, hostfxr_status>)funcHandle;
         hostfxr_handle outContextHandle;
         hostfxr_status status = funcDelegate(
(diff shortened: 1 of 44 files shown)


jasongin:
Linux tests are failing with the following error:

> Unhandled exception. System.DllNotFoundException: Unable to load shared library 'libdl' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable: 
/usr/share/dotnet/shared/Microsoft.NETCore.App/7.0.3/libdl.so: cannot open shared object file: No such file or directory
/home/runner/work/napi-dotnet/napi-dotnet/out/bin/Release/TestCases/napi-dotnet/net6.0/linux-x64/libdl.so: cannot open shared object file: No such file or directory
/usr/share/dotnet/shared/Microsoft.NETCore.App/7.0.3/liblibdl.so: cannot open shared object file: No such file or directory
/home/runner/work/napi-dotnet/napi-dotnet/out/bin/Release/TestCases/napi-dotnet/net6.0/linux-x64/liblibdl.so: cannot open shared object file: No such file or directory
/usr/share/dotnet/shared/Microsoft.NETCore.App/7.0.3/libdl: cannot open shared object file: No such file or directory
/home/runner/work/napi-dotnet/napi-dotnet/out/bin/Release/TestCases/napi-dotnet/net6.0/linux-x64/libdl: cannot open shared object file: No such file or directory
/usr/share/dotnet/shared/Microsoft.NETCore.App/7.0.3/liblibdl: cannot open shared object file: No such file or directory
/home/runner/work/napi-dotnet/napi-dotnet/out/bin/Release/TestCases/napi-dotnet/net6.0/linux-x64/liblibdl: cannot open shared object file: No such file or directory
   at Microsoft.JavaScript.NodeApi.JSNativeApi.Interop.dlopen(IntPtr fileName, Int32 flags)
   at Microsoft.JavaScript.NodeApi.JSNativeApi.Interop.Initialize(IntPtr libraryHandle) in /home/runner/work/napi-dotnet/napi-dotnet/src/NodeApi/Native/JSNativeApi.Interop.cs:line 36

Edit: Fixed by creating a symbolic link for `libdl.so` in `build.yaml`.

src/NodeApi/DotNetHost/HostFxr.cs
```diff
@@ -154,25 +154,49 @@ public unsafe struct hostfxr_initialize_parameters
...
+        [MarshalAs(UnmanagedType.FunctionPtr)] out load_assembly_and_get_function_pointer function)
+    {
+        nint funcHandle = NativeLibrary.GetExport(Handle, nameof(hostfxr_get_runtime_delegate));
+        var funcDelegate = (delegate* unmanaged[Cdecl]<
+                hostfxr_handle, hostfxr_delegate_type, nint*, hostfxr_status>)funcHandle;
+        nint outFunction;
+        hostfxr_status status = funcDelegate(hostContextHandle, delegateType, &outFunction);
+        var outFunctionDelegate = (delegate* unmanaged[Cdecl]<
+            byte*, byte*, byte*, nint, nint, nint*, hostfxr_status>)outFunction;
+        function = status != hostfxr_status.Success ? default! :
```
vmoroz, line 187: >function [](http://example.com/codeflow?start=8&length=8)

I wonder if we can make this parameter type nullable with the NotNullWhen(false). It should say that it is not null when returned hostfxr_status is 0.

vmoroz (approved):
:shipit:

src/NodeApi/DotNetHost/HostFxr.cs
jasongin: NotNullWhen only works with boolean return type.

PR #44, opened 2023-03-22
Stream marshalling

jasongin:
Marshal .NET `Stream` to/from Node.js `Duplex` / `Readable` / `Writable`.

 - Improve the `JSContext.Import()` method so that it supports importing from modules via `require()`.
    - There is no way to obtain the `require()` function via `napi_` APIs. The workaround is to pass it as a global by setting `global.require = require` before loading the module. For AOT this was the only workable solution I could find, since `require()` may need to be called during module registration, before it would be possible for JS code to pass it in via a property/method exported from the module.
 - Add a `baseClass` parameter to `JSClassBuilder<T>.DefineClass()`, so now it is possible to extend JS classes with new classes defined by .NET.
 - Add `NodeStream` class that implements .NET Stream APIs over a Node.js stream, and supports generating JS stream proxies for .NET streams. Support for stream subclass exports enables invoking .NET stream constructors from JS, e.g. `new FileStream(fileName)`.
 - Support constructor overloads in generated module exports code. (They were already supported for dynamic invocation.)
 - Misc improvements to the type-definitions generator: error reporting, doc-comment formatting, and more.


src/node-api-dotnet/index.js
```diff
@@ -11,6 +11,10 @@ const runtimeIdentifier = getRuntimeIdentifier();
 const nativeHostPath = __dirname + `/${runtimeIdentifier}/${assemblyName}.node`;
 const managedHostPath = __dirname + `/${targetFramework}/${assemblyName}.DotNetHost.dll`
 
+// The Node API module may need the require() function at initialization time; passing it as
+// a global is the best (only?) solution, since it cannot be obtained via any `napi_*` function.
+global.require = require;
```
vmoroz, line 16: It does not look good that we change the global.
Could we instead push the 'require' variable into the module after the its initialization?

README.md
```diff
@@ -136,6 +136,9 @@ convert values beteen JavaScript and C#. Here's a general summary of conversions
   - Custom marshaling and marshaling hints [may be supported later](
     https://github.com/jasongin/napi-dotnet/pull/25).
 
+### Stream across .NET and JS
```
vmoroz, line 140: >### [](http://example.com/codeflow?start=0&length=3)

nit: could we limit the length of strings?

vmoroz (approved):
:shipit:

README.md
jasongin: Can you clarify the suggestion? I'm confused because this is about streams, not strings.

src/node-api-dotnet/index.js
jasongin: I don't like it either. I spent some time trying to find any possible way to get `require` via any `napi_*` function, but nothing worked. We should give that feedback to the Node API working group.

For AOT modules, after initialization can be too late if the module exports some API that references a `Stream`, because setting up the module exports may involve invoking `require()`.

For .NET runtime modules, it is possible to pass in `require` via the host initialization function. I had actually implemented it that way first, but then changed it to be consistent with AOT. But if we expect AOT usage to be much less common then we can at least avoid using the global for non-AOT modules.

We could change AOT modules so that they also have an explicit initialization function. That would mean you can no longer just do a normal `require('MyModule.node')` to load an AOT module. Do you have any thoughts about that?

jasongin: I will merge this as it is now, then follow up later with something a little better.

README.md
jasongin: Fixed.

